PR #18, opened 2015-12-31
Also check for 'zfs receive' alias

jjlawren:
When checking the currently running processes for zfs send & receive activity, also check the `recv` alias.


jimsalterjrs:
Thanks!  Was unaware that "recv" alias was a thing.

PR #19, opened 2015-12-31
Use UNIX-style 'ps' options everywhere for compatibility

### LINKED ISSUE #17: Issues with 'ps' on SmartOS
It seems the assumptions for `ps` arguments don't work well on SmartOS or perhaps other Solaris-ish / Illumos-based systems.

When using the BSD-style options (grouped arguments without '-' preceding them), it falls back to the compatibility `/usr/ucb/ps` binary, which doesn't support the '-o' argument.

It looks like this can be solved (for SmartOS) by always using the UNIX-style options with a preceding hyphen and the '-Ao' flag instead of 'axo' for checking ZFS activity.

jjlawren:
Hopefully fixes #17 without the need for OS compatibility checking. Tested `ps` options on SmartOS, Ubuntu 14.04, and FreeBSD 10.2. The FreeBSD man page does not document the '-A' option, but the binary supported it with identical output.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -928,7 +928,7 @@ sub checklock {
 		return 2;
 	}
 
-	open PL, "$pscmd p $lockpid -o args= |";
+	open PL, "$pscmd -p $lockpid -o args= |";
 	my @processlist = <PL>;
 	close PL;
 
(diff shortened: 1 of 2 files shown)


jimsalterjrs:
also confirmed on Ubuntu 12.04 LTS.  Thank you for the pull request!

PR #26, opened 2016-03-08
Added -sshkey arg for automated cron runs

tmlapp:
Added -sshkey arg to to syncoid for unattended cron jobs. Adjust $sshcmd based on presence of the key.


jimsalterjrs:
Thanks for the code!  Note: spec file looks reasonable from here, but I am not an RPM expert and cannot really guarantee its quality or maintenance.

PR #77, opened 2017-04-13
added missing isroot args

AndCycle:
I just find out there are some extra sudo execute under root,

minor fix.


jimsalterjrs:
bugfix explicitly passes $sourceisroot and $targetisroot to getsnaps() invocations to avoid unnecessary use of sudo. Thank you @AndCycle

PR #92, opened 2017-06-20
Fix #91: do not use exit codes > 255

### LINKED ISSUE #91: syncoid: Do not use exit codes > 255
There is one error case in which syncoid return the exit code 256. Unfortunately, when exit codes > 255 are used, bash treats that as `exitcode % 256`. Since `256 % 256` is 0, this is treated as a successful exit.

Due to this problem, I've had failing syncs for a few weeks without noticing it, since bash always treated the syncoid invocation as successful, not triggering my error handling.

rschaeuble:
Exit codes > 256 are treated by bash as `exit_code % 256`, which is confusing. Furthermore, using code 256 is particularly problematic, as 256%256=0, which means success, and is not was was intended here.

For further information about special exit codes, see http://tldp.org/LDP/abs/html/exitcodes.html.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -764,7 +764,7 @@ sub getmatchingsnapshot {
 		print "                replication must be to a NON EXISTENT DATASET, which will\n";
 		print "                then be CREATED BY the initial replication process.\n\n";
 	}
-	exit 256;
+	exit 3;
 }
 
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
File this under "whoops!" Thank you.

PR #94, opened 2017-06-28
Update INSTALL to reflect crypto changes

tomposmiko:
In v1.4.6 Xenial and newer FreeBSD changes was added.
Change INSTALL to reflect them so it cannot mislead users about requirements.


jimsalterjrs:
accepted. I might need to actually remove the bits that try to force a crypto algorithm when none is specified; at one point the default algo for SSH was a real CPU hog and would severely slow down local transfers, but that hasn't been the case with the newer defaults lately.

tomposmiko:
thanks!

PR #98, opened 2017-07-04
add --sshcipher option

tomposmiko:
Add --sshcipher option.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -45,6 +45,9 @@ my $lscmd = '/bin/ls';
 if ( $args{'sshport'} ) {
   $sshport = "-p $args{'sshport'}";
 }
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
OK, sorry, not accepting this one because you're putting it in the wrong place.  There _is_ already a variable that sets the SSH cipher up at the top of syncoid; that's going to be the right place to set the cipher option if used.  Also, for intuitive use with people already familiar with the -c argument of ssh, I'm going to just make it -c here in syncoid as well.  I'm about to push my own commit for this; please update your own fork with that instead.

Thank you for the PR though, and sorry about the confusion!

tomposmiko:
No problem at all!
Take this one and my future PRs as a POC.
Thanks!

PR #121, opened 2017-08-03
Explicit LXD support

efoverskov:
I've hacked together some code to support LXD container snapshotting in Sanoid. It uses the lxc client utility to do the snapshots instead of zfs utility to avoid confusing LXD. First time using GitHub and very green in perl, so please bear with me.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -212,7 +212,7 @@ sub prune_snapshots {
 				# print "snap $path\@$snap has age $snaps{$path}{$snap}{'ctime'}, maxage is $maxage.\n";
 				if ( ($snaps{$path}{$snap}{'ctime'} < $maxage) && ($numsnapsthistype > $minsnapsthistype) ) {
 					if ( $islxd ) {
-					    push(@prunesnaps, "$lxc delete $section/snap");
+					    push(@prunesnaps, "$lxc delete $section/$snap");
 					} else {
 					    push(@prunesnaps, "$zfs destroy $path\@$snap");
 					}
@@ -523,7 +523,6 @@ sub getsnaps {
 	foreach my $snap (@rawsnaps) {
 		my ($fs,$snapname,$snapdate) = ($snap =~ m/(.*)\@(.*ly)\s*creation\s*(\d*)/);
 		my ($snaptype) = ($snapname =~ m/.*_(\w*ly)/);
-#		if ($snapname =~ /^autosnap/) {
 		if ($snapname =~ /(^snapshot-)*autosnap/) {
 			$snaps{$fs}{$snapname}{'ctime'}=$snapdate;
 			$snaps{$fs}{$snapname}{'type'}=$snaptype;
--- a/sanoid
+++ b/sanoid
@@ -524,7 +524,10 @@ sub getsnaps {
 		my ($fs,$snapname,$snapdate) = ($snap =~ m/(.*)\@(.*ly)\s*creation\s*(\d*)/);
 		my ($snaptype) = ($snapname =~ m/.*_(\w*ly)/);
 		if ($snapname =~ /(^snapshot-)*autosnap/) {
-			$snaps{$fs}{$snapname}{'ctime'}=$snapdate;
+		        if ($snapname =~ /^snapshot-/) {
+			        $snapname = substr $snapname, 9;
+			}
+		        $snaps{$fs}{$snapname}{'ctime'}=$snapdate;
 			$snaps{$fs}{$snapname}{'type'}=$snaptype;
 		}
 	}
--- a/sanoid
+++ b/sanoid
@@ -18,6 +18,7 @@ my %args = getargs(@ARGV);
 my $pscmd = '/bin/ps';
 
 my $zfs = '/sbin/zfs';
+my $lxc = '/usr/bin/lxc';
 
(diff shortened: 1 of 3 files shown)


jimsalterjrs:
Can you give me a human-readable overview of how your patch is intended to be used, please?

I think I grok it, but I'd like to get an idea of what it accomplishes, under what conditions, so I can better evaluate it and test it before accepting into master.

Thank you!

efoverskov:
I can sure try :)
When running LXD containers using ZFS backend, LXD has the capability to
utilize snapshots. However, snapshots made without the knowledge of LXD are
"invisible" to it, and unsupported. So, my idea was to simply use the LXD
client, confusingly named lxc, to do the snapshots.whenever a section in
sanoid.conf contains lxd = yes. I've had to work around LXD insisting on
prefixing the snapshot names it creates with "snapshot-", this behaviour is
hard coded. I've tested on one of my own systems, and both LXD and regular
zfs snapshots and pruning seem to work great. Couple of things still left
to test are monitor and syncoid. Also, recursion won't work at the moment,
I've added a few comments to the sample sanoid.conf.

I hope this explains what I'm trying to do. Like I said, it is functional
at present, but if it's good enough to include I don't know. But I've seen
a few threads requesting where people would like to use sanoid with LXD,
e.g. https://github.com/lxc/lxd/issues/3250. That's what prompted me to
attempt it, just love sanoid!


2017-08-04 19:48 GMT+02:00 Jim Salter <notifications@github.com>:

jimsalterjrs:
Thank you! I'd love to work with you on this; I'm very interested in `lxd` and it's been on my list of stuff to play with "real soon now" for quite a while.

If recursion isn't working yet, that's going to be a problem for me.  We may need to do a separate branch for your stuff until it's a little more mature. I'd really, really like to get lxd support clean and functional though.

Do you know whether it's possible to store your lxd containers in arbitrary datasets, btw? Another thread was opened today regarding possibly using wildcards to define module paths specifically for somebody who wants to do different things with datasets matching the glob `*-test`, but I'd really rather not do that.

As an lxd admin, is it possible to put your guest containers under, say, `guests/test/guest1` and `guests/test/guest2`, rather than `guests/guest1-test` and `guests/guest2-test` or similar?

Also: can we maybe look into *how* lxc knows a snapshot or dataset is "one of its own"? It might (or might not!) be cleaner to notify lxd of the datasets'/snapshots' existence and relation to it, rather than branching out to a different utility with potentially different arguments.

efoverskov:
Sounds great! For me, LXD is really great, almost as simple to work with as "real" vms, but with the light footprint of containers. Combined with zfs and sanoid, it just doen't get any better! Can't recommend it enough.

Regarding the recursion, I totally get that. That needs to work. Shouldn't be that difficult (Whew... Thanks [perldoc](https://perldoc.perl.org/). 

It is possible to use more than one dataset to store containers, and that could very well be a better route than introducing wildcards, which could prove to be a can of worms I bet. In LXD there's the concept of a storage pool, which can be e.g. a zfs dataset. They still don't afaik have an easy way to move containers between datasets on the same host, which is a bit odd, but there are workarounds. A bit odd, but maybe coming down the line.

As to the idea of notifying LXD about snapshots, I've had a look at the source for LXD (written in Go, so not exactly familiar to me). Based on that (well, that and Google) I think the issue is that LXD keeps track of the snapshots in its internal database. Remember, LXD supports other storage backends such as btrfs and LVM which also do snapshots, which must be why they need to do this. So probably not the way to go I think. Besides, the command line interface to LXD is quite simple and consistent, and seems to be fairly stable between versions.

So, I'll definitely have a look at what needs to be done to enable recursion to work, and hopefully whip up a proposal.

efoverskov:
Must admit, that if you're a lightweight, I'm sure to be a featherweight... I have no idea what would be best, but I'm thinking the safest would be for me to work in my own repo for now, that way we're sure that I can't hurt anyone but myself. But if you can figure out another way to do it that would make sense, I'm game. Haven't had time to look further at recursion, but everything still seems to work as expected my test system

thirtythreeforty:
I'll chime in here about Git if you don't mind. @efoverskov the best thing for you to do is to keep your work in your fork.  As you push new commits, they will appear in this pull request.  @jimsalterjrs there's no way for you to give @efoverskov access to *just* one branch—it's all or nothing.  That said, you can check this pull request out locally with [some Git-fu](https://help.github.com/articles/checking-out-pull-requests-locally/).

efoverskov:
Thanks for chiming in @thirtythreeforty . Sounds like the way to go  @jimsalterjrs do you agree?

jimsalterjrs:
Thank you for the chime, it's appreciated!


----------------

(Sent from my tablet - please blame any weird errors on autocorrect)

jimsalterjrs:
Hi @efoverskov - accepted quite a few PRs to syncoid and sanoid in the last few days, some of them relatively significant, at least one of which conflicts. You might want to update your fork =)

phreaker0:
Closing as this is stale and lxd support will probably not make it into sanoid, some of the functionality may be possible now with the pre/post script stuff #280

redmop:
Your best bet is to make a new bug request for this instead of using one that was closed in 2019.

PR #129, opened 2017-08-09
Don't die when recursively using no-sync-snap and there's a dataset with no snapshots

### LINKED ISSUE #127: Dataset with no snapshots breaks syncoid with --no-sync-snap
I've (attempted to) exclude a dataset from being snapshotted by sanoid, and from being replicated with syncoid (1.4.14)

```ini
[tank]
    use_template = default
    recursive = yes
[tank/excludeddataset]
    use_template = excluded
    recursive = yes

[template_excluded]
    autosnap = no
    autoprune = no
    monitor = no
```

However, syncoid, with the --no-sync-snap option, balks at the dataset, dying on: `CRIT: --no-sync-snap is set, and getnewestsnapshot() could not find any snapshots on source!`

(as an aside, it was difficult to work out which dataset this applied to - it's just not in the output anywhere, error message or otherwise - can this be amended?)

This corresponds to [this line](https://github.com/jimsalterjrs/sanoid/blob/32cd10097383c2568cf8d285a1a7f68167edd1c0/syncoid#L666), which I believe shouldn't be a `die` as changing it to a warn, as is a similar statement on [line 174](https://github.com/jimsalterjrs/sanoid/blob/32cd10097383c2568cf8d285a1a7f68167edd1c0/syncoid#L174), makes the script complete as expected.

Unless I'm doing something wrong with how I'm excluding datasets (can we have a provided excluded template as well?) this would seem to be a bug, as syncoid shouldn't be trying to copy a dataset with no snapshots, with the excluded template, anyway.

LordAro:
Fixes #127 

I'm afraid my editor is very keen on stripping trailing whitespace. I put it all in a separate commit just in case you don't like it :)

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1,6 +1,6 @@
 #!/usr/bin/perl
 
-# this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved 
+# this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
 # project's Git repository at https://github.com/jimsalterjrs/sanoid/blob/master/LICENSE.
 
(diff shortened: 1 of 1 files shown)


LordAro:
I question your git(hub) methods, but whatever. I'll let you close this when you're done

jimsalterjrs:
like your code, hate your pushy editor. =) 

implemented manually, thanks for the PR.

jimsalterjrs:
Do you have a suggestion as to how I could have handled it better? I'm open to it if you do; I'm a bit lightweight on git(hub).

I don't mind your editor stripping *trailing* whitespaces, but it also wanted to kill *leading* whitespaces on otherwise empty lines. Those are deliberate; they help me identify what is or isn't still part of a given sub.  In general, I really don't want to take a PR that *appears* to modify giant whacks of code that it really isn't changing; it makes it harder to identify issues that may have been introduced with that PR later.

LordAro:
You could've asked me to fix it! I could've quite easily rebased the branch (& pull request) to remove the commit entirely (one of the reasons why I separated it out, along with ease of diff viewing)

As for the leading whitespace, I guess that's your call. I just use the code indent level to work out what's where - the whitespace is just that - whitespace - I don't see it or anything, it just screws up where my cursor goes.

PR #130, opened 2017-08-09
Add example 'ignore' template

LordAro:
As discussed in #127

### DIFF
--- a/sanoid.conf
+++ b/sanoid.conf
@@ -54,17 +54,20 @@
 	monthly = 12
 	yearly = 0
 
-	### don't take new snapshots - snapshots on backup 
+	### don't take new snapshots - snapshots on backup
 	### datasets are replicated in from source, not
 	### generated locally
 	autosnap = no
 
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
implemented manually; I like the fix, but can't abide what your text editor wants to do with whitespace. thank you for the PR.

jimsalterjrs:
If all it was doing was eliminating trailing whitespace on actual lines of code, I'd consider it. Unfortunately, it's also removing LEADING whitespace on empty lines (which is therefore also trailing), which I really don't like because having it there helps me when arrowing around in the code, and also helps keep formatting consistent when pasting code blocks into text boxes online that support markdown.

I really, *really* don't want to accept cosmetic changes of hundreds of lines of code mixed in with functional changes to actual code in a single pull, also. It makes it harder to figure out what blew up, if something does blow up later on.  I'd consider a purely cosmetic PR related to whitespace all on its own... as long as, like I said, it didn't get rid of whitespace on otherwise empty lines, which I *do* actively want.

PR #131, opened 2017-08-10
Rewrite argument parser with GetOpt::Long and add helptext

### LINKED ISSUE #128: Feature request: Help text for the commands
Having to go back to the readme to find which flags each command supports is a bit tiresome, especially when it's incorrect (syncoid doesn't have a --verbose flag) or missing entirely (findoid)

Be good to have a -h/--help flag for all of the commands

LordAro:
How does this look? If you approve, I'll do syncoid & findoid as well

Fixes #128


jimsalterjrs:
Is this intended to be functional yet, or just sort of a demo of how things will go eventually?  I downloaded from your helptext branch and tried a simple `./sanoid --cron` and got a couple pages of errors.

LordAro:
Huh. It should've been working... Can you paste the errors?

LordAro:
Ah, I see what's happened. I added a `use warnings;` directive at the top of the file (copied across from my mini testing program), which has "revealed" a lot of warnings in the script.

I've updated my branch to remove that directive, try again?
It'd be good to have it included again at some point in the future, but that's not for this pull request :)

jimsalterjrs:
Well that's a little embarrassing - originally I wasn't even using strict; I was quite proud of myself a year or two ago when I updated sanoid and syncoid to function under strict. I *thought* strict actually *implied* warnings... obviously not.

I vaguely remember trying to work around the uninitialized value warnings and being frustrated that even sanitizing them with nz resulted in a warning about *nz* operating on an uninitialized value... but I think those memories are from another project entirely (a non-public one).

jimsalterjrs:
I added some ifdef checks to mute the majority of those warnings. Now the only ones left are the two "called too early to check prototype" for monitor_health and monitor_snapshots, which I frankly don't grok at first glance and don't have time for this morning. Any thoughts on those? (If not, I'll google and fix at some point; I just don't have the bandwidth right now.)

Also: are you using sanoid yourself? If so, are you dogfooding these changes? I haven't been able to figure out a good way to do automated testing without literally dedicating a machine to running dogfood versions of Sanoid in production and checking it to see if snaps are created and removed properly, etc.  So if you're not dogfooding this, I'll need to on a few of my own machines to QA the changes before accepting into master on my end. =)

LordAro:
The "called too early to check prototype" looks to be because the functions are called before they're declared in the file (like C code). Move the "main" stuff to the bottom of the file and those should be fixed fine

I am indeed using sanoid myself (hence all the recent stuff!) although I will admit I haven't made use of the above stuff with sanoid itself (which meant i missed all the warnings). But if some of it works (--cron) I see no reason why the rest of it wouldn't work either

Some sort of automated testing would be nice, but as you say, tricky to do without stubbing out all of zfs - you'd need to do full-on integration testing to get any reasonable results. I wonder if you could get ZoL running on one of the popular CI services... (travis-ci, circle-ci, etc)

jimsalterjrs:
Yeah, `--cron` calls the majority of `sanoid`'s functionality.  Ideally I'd need to also check on `--monitor-health` and `--monitor-snapshots`... *and* make sure snapshots are actually being taken and expired properly, and that health and freshness are being monitored properly, not just "don't see any errors, doot de do." =)

Are you willing to dogfood this branch for a bit, since it's mostly your changes in the code?

jimsalterjrs:
> The "called too early to check prototype" looks to be because the functions are called before they're declared in the file (like C code). Move the "main" stuff to the bottom of the file and those should be fixed fine

Hrm. Shouldn't it be issuing a warning for *every* function, then, since they're *all* declared after main? What am I missing here?

LordAro:
Yeah, I can use this branch for a bit

Your guess is as good as mine for why it's only warning for the 2 functions though. monitor_health() and monitor_snapshots() aren't the only functions called in main... 
Regardless, I feel it'd be good to move the "main" code to the bottom of the file anyway - it'll be closer to the helptext and so more obvious that both should be updated

LordAro:
Lots of juicy changes to syncoid. See commit message for more details

jimsalterjrs:
Hey @LordAro, can you confirm that you've been dogfooding these changes with no issues for a couple weeks now?  If so, I'll go ahead and accept the PR. I appreciate the work!

LordAro:
Ack, I'm afraid it seems I haven't actually been using this. The backup setup has stabilised now (i'm no longer coming in every morning to find lots of icinga "backups out of date emails"!) so I've not been focusing on it

I do however have a little bit more work to do on it, I'll try to make use of it then

LordAro:
Good call on actually making me test it ;)

jimsalterjrs:
Let me know how we're doing on dogfooding, please. =)

LordAro:
As far as I can tell, everything that I'm touching works :)

I would strongly advise taking a look through the changes yourself though, if I've done something else silly like invert the meaning of "recursive" ;)

LordAro:
Rebased. I'm happy that this is working now.

LordAro:
I could do findoid, but i'm not actually convinced anyone uses it ;)

redmop:
I think it's just me. Proxmox and LXC. I don't use it often.



Shawn T Perry
Red Mop Computing Services
phone: 303-881-2623
email: shawn@redmop.com

LordAro:
@jimsalterjrs any issues from you?

jimsalterjrs:
Looks good from here. Merged.

jimsalterjrs:
Hey @LordAro, I had to revert back to the old syncoid. Your compressargset routine returns a hash reference into a string, which the rest of the codebase expects to be, well, a string. Can you fix, verify that syncoid works using a remote source or target both with or without compress arguments set, and then we can merge again?

PR #141, opened 2017-08-31
Update RPM Spec

dcrdev:
This pull request bumps the version number within the rpm spec file, additionally it introduces the addition of systemd timers. This is convenient because it means the user can activate sanoid out of the box via "systemctl enable sanoid.timer".

By default the timer is set to run every minute, but this can be overridden by copying the unit and timer into /etc/systemd/system and making the appropriate changes.

The systemd timers are entirely optional within the spec itself, they can be toggled on/off with the _with_systemd variable.

### DIFF
--- a/sanoid.spec
+++ b/sanoid.spec
@@ -110,7 +110,7 @@ echo "* * * * * root %{_sbindir}/sanoid --cron" > %{buildroot}%{_docdir}/%{name}
 %endif
 
 %changelog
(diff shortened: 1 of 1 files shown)


dcrdev:
Test packages available here:
https://repo.dcrdev.com/el7/x86_64/sanoid-1.4.14-2.el7.centos.noarch.rpm
https://repo.dcrdev.com/el7/SRPMS/sanoid-1.4.14-2.el7.centos.src.rpm

PR #148, opened 2017-09-13
Build sanoid as a native Debian package

DHowett:
**Submitting for Discussion**

Hi @jimsalterjrs! I've done the work necessary to bootstrap sanoid as a native Debian package. It's a little rough right now, and I've made a few concessions:

* `CHANGELIST` is partially duplicated as `debian/changelog` in the debian changelog format
* Version numbering is handled by `debian/changelog` and, optionally, the `dch` tool
* The package requires `zfsutils-linux`
* The systemd timer fires every 15 minutes -- it's not very fine-grained :smile:

### Building the package

```
$ dpkg-buildpackage -uc -us
  # -uc = unsigned changelog
  # -us = unsigned sources
```

I'm not a Debian maintainer, so I can't speak to how conformant this package will be if you build and submit it! I'd therefore like to get a discussion going about what it needs to be a good representation of your project.

Thanks!


debian/control
```diff
@@ -0,0 +1,14 @@
+Source: sanoid
+Section: unknown
+Priority: optional
+Maintainer: Dustin L. Howett <dustin@howett.net>
```
DHowett: Whoops! I meant to purge all references to my name, as I'm probably not going to be the maintainer if this merges

debian/control
```diff
@@ -0,0 +1,14 @@
...
+Build-Depends: debhelper (>= 9)
+Standards-Version: 3.9.8
+Homepage: https://github.com/jimsalterjrs/sanoid
+Vcs-Git: https://github.com/jimsalterjrs/sanoid.git
+Vcs-Browser: https://github.com/jimsalterjrs/sanoid
+
+Package: sanoid
+Architecture: all
+Depends: ${misc:Depends}, ${perl:Depends}, zfsutils-linux, libconfig-inifiles-perl
```
DHowett: Other perl packages might have to go here; `perl:depends` expands to `perl`, really, so the package gets a perl dependency.

debian/control
```diff
@@ -0,0 +1,14 @@
...
+Description: Policy-driven snapshot management and replication tools
```
DHowett: there's an opportunity for a longer description here, it'll just have to be indented by one space so the control parser doesn't eat it

debian/copyright
```diff
@@ -0,0 +1,33 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: sanoid
+Source: <https://github.com/jimsalterjrs/sanoid>
+
+Files: *
+Copyright: 2017 Jim Salter <github@jrs-s.net>
+License: GPL-3.0+
+
+Files: debian/*
+Copyright: 2017 Jim Salter <github@jrs-s.net>
```
DHowett: (Since the changes in the debian/ directory are small, I'm comfortable assigning copyright to you)

debian/sanoid.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Snapshot ZFS Pool
+Requires=zfs.target
+After=zfs.target
+ConditionFileNotEmpty=/etc/sanoid/sanoid.conf
```
DHowett: ("only start if sanoid.conf has been populated")

debian/control
phreaker0: can you replace "zfsutils-linux" with "zfsutils-linux | zfs", then it would also work with custom debian packages generated from the zfs git repository

debian/copyright
jimsalterjrs: Thanks for your work on this!

debian/control
jimsalterjrs: @phreaker0 , I made your change - please review and make sure syntax is appropriate if you have a system which would benefit from it; I'm a complete newb with debian packaging.

UltraPhil:
Uou have no idea how much this would make my day if sanoid(+syncoid) was integrated in the debian/ubuntu official repositories.  Think we could get a sponsor developper from debian/ubuntu to integrate officially?

debian/control
phreaker0: thx, i verified that the generated sanoid package works for both the official zfs packages provided by debian/ubuntu and the custom debian packaged generated from the official zfs git repository

PR #157, opened 2017-11-07
implemented frequent snapshots with configurable period

### LINKED ISSUE #75: Taking snapshots more frequently than once every hour
Would be nice to be able to manually specify the number of minutes

phreaker0:
### Description

Adds option to take more frequent snapshots because hourly ones are not enough in some cases. The used period is configurable in the range of 1 - 30 minutes, but should divide an hour without remainder.

### How Has This Been Tested?

Only limited testing by me in a VM.

Fixes #75


phreaker0:
thx for the feedback, i will address it and update the PR

kpaskett:
I grabbed this and am using it. Works great.
The one issue I had was that I tried to override frequent_period in the sanoid.conf and got an error. I may have mis-typed something though. I haven't had time to go back and verify.

phreaker0:
@kpaskett: you probably got something like this:
`FATAL ERROR: I don't understand the setting frequent_period you've set in [storage] in /etc/sanoid/sanoid.conf.`

This is caused if the default sanoid configuration doesn't have the new "frequent_period" set. This will be fixed by hardcoding defaults.

jimsalterjrs:
@phreaker0 , I apologize: I misled you earlier about the hardcoded defautls thing. I USED to do that, but after a LOT of internal struggle with it, decided to STOP doing it because all possible values should be in sanoid.defaults.conf. My idea here was that sanoid.defaults.conf should be, effectively, a complete set of documentation as to everything you can do inside sanoid.conf.

I think the proper way forward here is to make Sanoid check for a minimum version number set in sanoid.defaults.conf; something that isn't currently implemented.  So let's say that the current prod sanoid.defaults.conf is 1.0.0, and the version of sanoid.defaults.conf that you're using with frequent_period set is 1.0.1.  Now when you invoke sanoid, sanoid has a minimum_defaults_version of 1.0.1, and instead of getting the generic "I don't understand that parameter" error, we get a "FATAL ERROR: you're using sanoid.defaults.conf v1.0.0; this version of Sanoid requires a minimum sanoid.defaults.conf v1.0.1".

What do you think? This is hopefully something that shouldn't come up very often, but it's something we should definitely handle RIGHT since it's a game-breaker for people who mismatch their updates. =)

phreaker0:
@jimsalterjrs did you see my last change? I only set the hardcoded defaults if they are not already set in the sanoid.defaults.conf. As frequent snapshots are disabled by default an upgrade should work as before. This is nice for the end user who doesn't need to adapt the default configuration file but if more configuration options are added they also have to be checked and this legacy handling will grow which probably isn't nice.

The obvious downside of your proposed approach is that an upgrade (manually or automatic via package repositories) will break sanoid snapshots if there are new features with new configuration options, even if they are not used, which is not very nice. This would be even worse if the end user doesn't check if sanoid errors out and the snapshots just stopped being taken. I can imagine the later happening on desktop systems where no email sending is configured for cron/systemd timer errors.

jimsalterjrs:
I haven't looked at it yet, I'm having an "interesting" day with lots of distractions.

Automatic upgrades (which I have not actually designed) should automatically upgrade sanoid.defaults.conf, too - it says very clearly in its header:

````
###################################################################################
# default template - DO NOT EDIT THIS FILE DIRECTLY.                              #
#                    If you wish to override default values, you can create your  #
#                    own [template_default] in /etc/sanoid/sanoid.conf.           #
#                                                                                 #
#                    you have been warned.                                        #
###################################################################################
````

I'm not SUPER concerned about somebody who never so much as runs `sanoid --cron` to check everything's okay after a manual upgrade. I mean, it would be kinda nice to catch them somehow in a way that does not mean snapshots not being taken, but I don't know that it's worth making the code less clean.

This is only currently an issue because there IS no automatic upgrade process (which should automatically upgrade sanoid.defaults.conf along with the executables), and because it has very rarely needed any upgrading since I've tried to keep the conf file syntax as stable as possible.

Thoughts?

phreaker0:
Didn't thought about the automatic upgrade/replacement of the default configuration files in the case of packages, this prevents the mentioned downside.
Let's go with your approach.

phreaker0:
@jimsalterjrs i implemented your proposed approach. After the configuration check the version section is removed so it one interfere later.

Please also check out the new comments regarding the frequent_period.

ahjohannessen:
@jimsalterjrs Is this going anywhere?

jimsalterjrs:
Sorry folks I haven't reviewed it yet. I'm down with the flu right now, so no fevered code reviewing. :) I will look into it next week and go from there.

phreaker0:
@jimsalterjrs forget my last comment, i think this PR should be kept as it is now (+ requested changes by you if there are any) as it is running fine for me. I use it in production on several (10+) machines and works as expected and i also wrote a simple test which verifies this in a vm by taking snapshots over a whole year.
So review it please when your time allows it :-)

rwarren:
+100 on this!  1 hr resolution is a bit limiting. We're going to check out this fork. Thanks, @phreaker0 !

jimsalterjrs:
@phreaker0 can you resolve conflicts please?

phreaker0:
done :-)

phreaker0:
@jimsalterjrs 
one year of birth is approaching :-)

PR #161, opened 2017-11-08
removed trailing whitespace

phreaker0:
next try :-) i looked through all the open pull requests and they all need to be rebased, so maybe we can remove the whitespaces now, but i am also fine to do it later.


LordAro:
I tried that before in #129 , good luck :p

PR #174, opened 2017-12-06
implemented a simple test which will take snapshots over a whole year…

phreaker0:
… and checks the resulting snapshot list

The test is intended to be ran in a dedicated VM as it will mess with the system time to properly test
sanoid. A simple check is in place to make sure it is running in a vm, if not it asks the user if he wan't to proceed.
It is very time consuming but I use it to test my changes and it really help's a lot with making sure nothing broke. And so it may be useful for others too.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,7 @@
+sanoid (1.4.17-SNAPSHOT) unstable; urgency=medium
+
+-- Jim Salter <github@jrs-s.net>  Wed, 9 Aug 2017 12:28:49 -0400
+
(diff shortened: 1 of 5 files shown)


jimsalterjrs:
thank you for this! I've been kinda stressing out over having no automated tests for a while now. I haven't had a chance to really look at this yet, but it's obviously not going to break anything for normal users so I'm going to go ahead and merge. I might want to chat with you some about usage later. I'd LOVE to get automated continuous integration set up here.

PR #180, opened 2017-12-21
Adds support for zfs send -c, and a few other efficiency improvements for special cases.

ckeasling:
These are a series of commits intended to make initial synchronization of largeish datasets more efficient.  It also adds support for the recently-released send compressed mode (well, recently on Linux with zfsonlinux-0.7, I don't know about other zfs distributions).  In addition, a bug that (probably?) would result in use of undefined values when the --no-command-check option was resolved.

New options:
--compress=zfs adds "-c" to "zfs send", to enable sending compressed data without the need to decompress from the pool, compress for transfer, decompress on receive, then compress again when writing to the pool.
--direct bypasses mbuffer, pv, and external compression, and avoids checking these things.  My motivation was driven by the WARN messages emitted when these weren't found.  Since there is no way to disable those messages (i.e., --quiet doesn't silence them), and since they aren't needed in my use case, I figured an option to bypass them was suitable.
--rsh goes old-school and eliminates the ssh encryption overhead.  An ssh which permits -c none would be preferable, but isn't readily available (it's not available in vanilla ssh on RHEL, and installing alternate ssh packages seems icky).  Certainly using rsh isn't suitable in many cases, but it saves quite a bit of time transferring between two systems on the same switch.

These changes are fairly minimal and add some helpful functionality to my world.  Please advise if they are inadequate.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -82,6 +82,8 @@ if (!defined $targethost) { $targethost = ''; }
 my $sourcesudocmd = $sourceisroot ? '' : $sudocmd;
 my $targetsudocmd = $targetisroot ? '' : $sudocmd;
 
+my $send_compressed = $args{'compress'} eq 'zfs' ? ' -c ' : '';
+
 # figure out whether compression, mbuffering, pv
 # are available on source, target, local machines.
 # warn user of anything missing, then continue with sync.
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
Hi @ckeasling. Sorry this one sat for so long! I'm not willing to accept --rsh, but if you can resolve conflicts with the --compress=zfs option, I'll happily accept that one. Rather than your proposed --direct, just go ahead and add suppression of the WARN messages you mentioned with --quiet; I'm okay with assuming someone using --quiet is sure of what they're looking for, at least in that specific case!

jimsalterjrs:
closed due to inactivity

PR #181, opened 2017-12-21
Zfs Recursion

jMichaelA:
Allows the user in the config to specify to use zfs snapshot recursion option other than using Perl to recursively take snapshots. So your config would look like 
############################################
[pool4/zones/zone1]
        use_template = production
        recursive = zfs

[pool4/zones/zone2]
        use_template = production
        recursive = yes

zone1 would use use zfs recursion and zone2 would use Sanoid's way.
Zone1 recursive snapshot would look like: /sbin/zfs snapshot -r pool4/zones/zone1@autosnap_2017-12-19_15:37:21_hourly ....

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -503,12 +503,11 @@ sub take_snapshots {
 					%datestamp = get_date();
 					# print "we should have had a $type snapshot of $path $maxage seconds ago; most recent is $newestage seconds old.\n";
 
-					# use zfs recursion if specified in config
+					# use zfs (atomic) recursion if specified in config
 					if ($config{$section}{'zfs_recursion'}) {
-						push(@newsnaps, "-r $path\@autosnap_$datestamp{'sortable'}_$type");
-
-					}else{
-						push(@newsnaps, "$path\@autosnap_$datestamp{'sortable'}_$type");
+						push(@newsnaps, "$path\@autosnap_$datestamp{'sortable'}${dateSuffix}_$type\@");
+					} else {
+						push(@newsnaps, "$path\@autosnap_$datestamp{'sortable'}${dateSuffix}_$type");
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
@phreaker0 add this one to the list of "rebase or nah?"

jimsalterjrs:
@phreaker0 tag me when you've done as much rebasing as you want to do today, pls, and we'll finish up this merge spree and I'll get to testing so we can push the next release. Thank you as always!

phreaker0:
@jimsalterjrs i'm done, I assigned some PR's to their original authors for rebasing. I only need to update the "prepare for v1.4.19 release" PR adding the recent additions so you can go ahead and test, thanks :-)

jimsalterjrs:
after updating sanoid.defaults.conf and running sanoid --cron, I got a couple pages of this:

```
Use of uninitialized value in regexp compilation at /usr/local/bin/sanoid line 852.
Use of uninitialized value in regexp compilation at /usr/local/bin/sanoid line 847.

```

jimsalterjrs:
Relevant lines, 847-852:

```
                if (grep( /^$ini{$section}{'recursive'}$/, @istrue ) || grep( /^$ini{$section}{'skip_children'}$/, @istrue )) {
                        @datasets = getchilddatasets($config{$section}{'path'});
                        DATASETS: foreach my $dataset(@datasets) {
                                chomp $dataset;

                                if (grep( /^$ini{$section}{'skip_children'}$/, @istrue )) {
```

jimsalterjrs:
Looks like we need a preliminary check to see if `$ini{$section}{'skip_children'}` and possibly `$ini{$section}{'recursive'}` have been initialized?

phreaker0:
@jimsalterjrs thanks for catching this, will prepare a PR

jimsalterjrs:
@phreaker0 still got problems with this one... in testing I'm seeing 

````
Use of uninitialized value in pattern match (m//) at /usr/local/bin/sanoid line 866.
````

jimsalterjrs:
I think the issue here is that for datasets which don't have recursion set at all, $ini{$section}{'recursive'} needs to be initialized, perhaps to 'none'.

jimsalterjrs:
Actually there's also a problem that recursive is defined as a toggle in init(), and now... it's not a toggle anymore, there are three separate states. I'm reverting this entirely until it's a bit more thoroughly baked.

phreaker0:
sry for that, will look into it

Issue #182, opened 2017-12-24
quiet mode vs `INFO: deferring pruning`


martinvw:
Hi,

Thanks for your great scripts, they are an amazing help.

I have one question I used the `--quiet` option in a cron which runs every minute, at the same time I'm running an initial sync to a remote data center, which takes a little longer. 

However I received a lot info  / warnings past night, once per minute to be precise :-)

I think that either we should not print this statement (when running quiet) and otherwise we should print WARN instead. 

If you let me know what you think is best, I can make a PR, please also consider the the line below it:

https://github.com/jimsalterjrs/sanoid/blob/f6519c0aea4c624161f92a1304943d1c60e714bc/sanoid#L238

https://github.com/jimsalterjrs/sanoid/blob/f6519c0aea4c624161f92a1304943d1c60e714bc/sanoid#L247

Thanks!

---

The Log:

```
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-20_21:27:46_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-20_22:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-20_23:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_00:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_01:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_02:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_03:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_04:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
INFO: deferring pruning of kilimanjaro/media@autosnap_2017-12-21_05:00:01_hourly - kilimanjaro/media is currently in zfs send or receive.
[...]

PR #183, opened 2017-12-29
Made two INFO prints mutable

martinvw:
Added a quiet check for two non-controllable INFO prints

Fixes: #182


martinvw:
@jimsalterjrs did you notice this PR?

Issue #69, opened 2017-01-25
syncoid doesn't check for existence of parent dataset on target


mailinglists35:
i have this tree of datasets:

server
server/backups
server/backups/fortigate60d

attempting to directly sync the child fortigate60d hangs, because it checks that child does not exist on target, but it assumes the parent exists (which does not), and continues execution, resulting in a hang (mbuffer stalled in 100% cpu waiting for data)

imho it should check also for existance of target's parent after checking for existance of target, and iterate that check recursively up to the target root dataset

another offtopic observation i have is that getting the list snapshots can be done faster and more precisely this way
zfs list -t snapshot -o name -s creation -H -d1 server/backups/fortigate60d

```root@server:/usr/local/bin# syncoid --debug server/backups/fortigate60d backup-usb/backups/fortigate60d
DEBUG: checking availability of /usr/bin/lzop on source...
DEBUG: checking availability of /usr/bin/lzop on target...
DEBUG: checking availability of /usr/bin/lzop on local machine...
DEBUG: checking availability of /usr/bin/mbuffer on source...
DEBUG: checking availability of /usr/bin/mbuffer on target...
DEBUG: checking availability of /usr/bin/pv on local machine...
DEBUG: syncing source server/backups/fortigate60d to target backup-usb/backups/fortigate60d.
DEBUG: checking to see if backup-usb/backups/fortigate60d on  is already in zfs receive using  /bin/ps -Ao args= ...
DEBUG: checking to see if target filesystem exists using "  /sbin/zfs get -H name backup-usb/backups/fortigate60d 2>&1 |"...
DEBUG: getting list of snapshots on server/backups/fortigate60d using  /usr/bin/sudo /sbin/zfs get -Hpd 1 creation server/backups/fortigate60d |...
DEBUG: target backup-usb/backups/fortigate60d does not exist.  Finding oldest available snapshot on source server/backups/fortigate60d ...
DEBUG: getting estimated transfer size from source  using "  /sbin/zfs send -nP server/backups/fortigate60d@zfs-auto-snap_weekly-2016-11-20-0447 2>&1 |"...
DEBUG: sendsize = 109406392
INFO: Sending oldest full snapshot server/backups/fortigate60d@zfs-auto-snap_weekly-2016-11-20-0447 (~ 104.3 MB) to new target filesystem:
[...]

PR #189, opened 2018-01-05
added ability to skip datasets... simply set syncoid:no-sync=true

attie:
I had a use case where I _do not_ wish to backup large / scratch datasets, but I _do_ wish to backup the tree that they are in.

To skip a dataset, set `syncoid:no-sync` to `true`:

```
zfs set syncoid:no-sync=true tank/users/attie/scratch
```

You can subsequently backup `tank/users` without including `tank/users/attie/scratch`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,14 @@ If ZFS supports resumeable send/receive streams on both the source and target th
 
 As of 1.4.18, syncoid also automatically supports and enables resume of interrupted replication when both source and target support this feature.
 
+##### Syncoid Dataset Properties
+
++ syncoid:no-sync
+
+  Setting this to `true` will prevent the dataset from being handled by syncoid in _any_ way - it will be skipped. This can be useful for preventing certain datasets from being transferred when recursively handling a tree.
+
+  Note that this will also prevent syncoid from handling the dataset if given explicitly on the command line.
+
 ##### Syncoid Command Line Options
 
 + [source]
--- a/README.md
+++ b/README.md
@@ -140,7 +140,9 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
   _Note_: this will also prevent syncoid from handling the dataset if given explicitly on the command line.
 
-  _Note_: syncing a child of a no-sync dataset will currently result in a critical error
+  _Note_: syncing a child of a no-sync dataset will currently result in a critical error.
+
+  _Note_: empty properties will be handled as if they were unset.
 
 ##### Syncoid Command Line Options
 
--- a/README.md
+++ b/README.md
@@ -120,11 +120,27 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 ##### Syncoid Dataset Properties
 
-+ syncoid:no-sync
++ syncoid:sync
 
-  Setting this to `true` will prevent the dataset from being handled by syncoid in _any_ way - it will be skipped. This can be useful for preventing certain datasets from being transferred when recursively handling a tree.
+  Available values:
 
-  Note that this will also prevent syncoid from handling the dataset if given explicitly on the command line.
+  + `true` (default if unset)
+
+    This dataset will be synchronised to all hosts.
+
+  + `false`
+
+    This dataset will not be synchronised to any hosts - it will be skipped. This can be useful for preventing certain datasets from being transferred when recursively handling a tree.
+
+  + `host1,host2,...`
+
+    A comma seperated list of hosts. This dataset will only be synchronised by hosts listed in the property.
+
+    _Note_: this check is performed by the host running `syncoid`, thus the local hostname must be present for inclusion during a push operation // the remote hostname must be present for a pull.
+
+  _Note_: this will also prevent syncoid from handling the dataset if given explicitly on the command line.
+
+  _Note_: syncing a child of a no-sync dataset will currently result in a critical error
 
 ##### Syncoid Command Line Options
 
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
   + `host1,host2,...`
 
-    A comma seperated list of hosts. This dataset will only be synchronised by hosts listed in the property.
+    A comma separated list of hosts. This dataset will only be synchronised by hosts listed in the property.
 
     _Note_: this check is performed by the host running `syncoid`, thus the local hostname must be present for inclusion during a push operation // the remote hostname must be present for a pull.
 
--- a/syncoid
+++ b/syncoid
@@ -183,9 +183,20 @@ sub syncdataset {
 
 	if ($debug) { print "DEBUG: syncing source $sourcefs to target $targetfs.\n"; }
 
-	if (getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:sync') eq 'false') {
+	my $sync = getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:sync');
+
+	if ($sync eq 'true' || $sync eq '-') {
+		# definitely sync this dataset - if a host is called 'true' or '-', then you're special
+	} elsif ($sync eq 'false') {
 		if (!$quiet) { print "INFO: Skipping dataset (syncoid:sync=false): $sourcefs...\n"; }
 		return 0;
+	} else {
+		my $hostid = hostname();
+		my @hosts = split(/,/,$sync);
+		if (!(grep $hostid eq $_, @hosts)) {
+			if (!$quiet) { print "INFO: Skipping dataset (syncoid:sync doesn't include $hostid): $sourcefs...\n"; }
+			return 0;
+		}
 	}
 
 	# make sure target is not currently in receive.
--- a/syncoid
+++ b/syncoid
@@ -185,7 +185,8 @@ sub syncdataset {
 
 	my $sync = getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:sync');
 
-	if ($sync eq 'true' || $sync eq '-') {
+	if ($sync eq 'true' || $sync eq '-' || $sync eq '') {
+		# empty is handled the same as unset (aka: '-')
 		# definitely sync this dataset - if a host is called 'true' or '-', then you're special
 	} elsif ($sync eq 'false') {
 		if (!$quiet) { print "INFO: Skipping dataset (syncoid:sync=false): $sourcefs...\n"; }
--- a/syncoid
+++ b/syncoid
@@ -183,8 +183,8 @@ sub syncdataset {
 
 	if ($debug) { print "DEBUG: syncing source $sourcefs to target $targetfs.\n"; }
 
-	if (getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:no-sync') eq 'true') {
-		if (!$quiet) { print "INFO: Skipping dataset (syncoid:nosync=true): $sourcefs...\n"; }
+	if (getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:sync') eq 'false') {
+		if (!$quiet) { print "INFO: Skipping dataset (syncoid:sync=false): $sourcefs...\n"; }
 		return 0;
 	}
 
--- a/syncoid
+++ b/syncoid
@@ -184,7 +184,7 @@ sub syncdataset {
 	if ($debug) { print "DEBUG: syncing source $sourcefs to target $targetfs.\n"; }
 
 	if (getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:no-sync') eq 'true') {
-		print "Skipping dataset (syncoid:nosync=true): $sourcefs...\n";
+		if (!$quiet) { print "INFO: Skipping dataset (syncoid:nosync=true): $sourcefs...\n"; }
 		return 0;
 	}
 
--- a/syncoid
+++ b/syncoid
@@ -183,6 +183,11 @@ sub syncdataset {
 
 	if ($debug) { print "DEBUG: syncing source $sourcefs to target $targetfs.\n"; }
 
+	if (getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'syncoid:no-sync') eq 'true') {
+		print "Skipping dataset (syncoid:nosync=true): $sourcefs...\n";
+		return 0;
+	}
+
 	# make sure target is not currently in receive.
 	if (iszfsbusy($targethost,$targetfs,$targetisroot)) {
 		warn "Cannot sync now: $targetfs is already target of a zfs receive process.\n";


UltraPhil:
The doc should be updated too if the PR is approved.

attie:
Indeed, please see new commits.

attie:
I feel that some level of parent checking is required as per #69...

Example:
```
$ zfs get -r syncoid:no-sync test_src
NAME                                PROPERTY         VALUE            SOURCE
test_src                            syncoid:no-sync  -                -
test_src/users                      syncoid:no-sync  -                -
test_src/users/attie                syncoid:no-sync  -                -
test_src/users/attie/no_sync        syncoid:no-sync  true             local
test_src/users/attie/no_sync/child  syncoid:no-sync  false            local
```

`.../users/attie/no_sync/child` will fail, as `.../users/attie/no_sync` does not exist:

```
$ syncoid -r test_src/users test_dst/mirror/users
INFO: Sending oldest full snapshot test_src/users@syncoid_perdy-base_2018-01-06:10:52:24 (~ 9 KB) to new target filesystem:
41.5KiB 0:00:00 [3.88MiB/s] [==============================================] 436%
INFO: Sending oldest full snapshot test_src/users/attie@syncoid_perdy-base_2018-01-06:10:52:24 (~ 9 KB) to new target filesystem:
41.5KiB 0:00:00 [3.23MiB/s] [==============================================] 436%
INFO: Skipping dataset (syncoid:nosync=true): test_src/users/attie/no_sync...
INFO: Sending oldest full snapshot test_src/users/attie/no_sync/child@syncoid_perdy-base_2018-01-06:10:52:25 (~ 9 KB) to new target filesystem:
40.2KiB 0:00:00 [3.11MiB/s] [==============================================] 423%
cannot open 'test_dst/mirror/users/attie/no_sync/child': dataset does not exist
cannot receive new filesystem stream: dataset does not exist
CRITICAL ERROR:  /sbin/zfs send test_src/users/attie/no_sync/child@syncoid_perdy-base_2018-01-06:10:52:25 | /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null | /usr/bin/pv -s 9728 |  /sbin/zfs receive -F test_dst/mirror/users/attie/no_sync/child failed: 256 at /usr/local/bin/syncoid line 241.
```

I think that recursive syncs should check/require the root target to exist (as-per `cp` / `rsync`, etc...), but should create any other intermediaries as required.

An alternative could be to create the skipped dataset on the destination, but not include any of its data, which I think is less acceptable.

gregorydlogan:
Maybe I'm being dense, but what's the difference between creating intermediaries as required, and creating the skipped datasets but not populating the data?  Wouldn't these both end up with empty intermediate datasets?

attie:
It's subtle, but consider the following hierarchies (the existence of `test_src/users/attie/no_sync/child` changes):

```
$ zfs get -r syncoid:no-sync test_src
NAME                                PROPERTY         VALUE            SOURCE
test_src                            syncoid:no-sync  -                -
test_src/users                      syncoid:no-sync  -                -
test_src/users/attie                syncoid:no-sync  -                -
test_src/users/attie/no_sync        syncoid:no-sync  true             local
test_src/users/attie/no_sync/child  syncoid:no-sync  false            local
```

```
$ zfs get -r syncoid:no-sync test_src
NAME                                PROPERTY         VALUE            SOURCE
test_src                            syncoid:no-sync  -                -
test_src/users                      syncoid:no-sync  -                -
test_src/users/attie                syncoid:no-sync  -                -
test_src/users/attie/no_sync        syncoid:no-sync  true             local
```

If we were to only create parents when required, then:
* The first would have `.../users/attie/no_sync` created on the target, because it is needed by `.../child`.
* The second would not have `.../users/attie/no_sync` created on the target, because nothing requires that it exist.

IMO, only creating when necessary is a cleaner solution, but it's a pretty minor difference. (less work, reducded leak of dataset names, etc...)

phreaker0:
@attie a problem one has to be aware of with creation of empty intermediates on the target is, that in case sometime you want to also sync the excluded intermediate dataset, it won't work because they aren't the same and one has to destroy the whole tree to get them

Besides that I like the possibility to exclude datasets like you implemented it. But if using an id instead of just a boolean for the property value the functionality can be extended like imagine the following use case:

- replication to more than one target host
- some datasets should be excluded only for a specific one, some for all, ...
- with id's in the property this could than be utilized like:
-  syncoid --exclude="vms,cache" or syncoid --include="backups"

what do you think?

attie:
Thanks for your feedback @phreaker0.
I appreciate this the issue around "syncing a dataset later" and was part of my reservation... Perhaps it's okay to leave as-is for now?

Regarding your further comment about specifying *where* to backup a dataset, excluding from certain places is something that I've been thinking about.

We could extend the options for `syncoid:no-sync` values:
* `false` - sync everywhere
* `true` - sync nowhere
* `some:${NAME_1},${NAME_2}` - sync only to `${NAME_1}` and `${NAME_2}`

Maintaining backwards compatibility (I'm actively using this now), and adding the feature... how does that sound?

samarium:
With respect to non synced intermediate datasets and having to delete the entire subtree if you want to sync the non synced datasets, I would do something like the below.

Given datasets:
src/sync1/unsync2/sync3/sync4
dst/sync1/unsync2/sync3/sync4
where unsync2 is unsynced on dst:

zfs rename dst/sync1/unsync2/sync3 dst/sync1/unsync2_sync3
zfs send src/sync1/unsync2@latest | zfs recv -F dst/sync1/unsync2
zfs rename dst/sync1/unsync2_sync3 dst/sync1/unsync2/sync3

Since I've just done something like that, but using syncoid, the steps were slightly different for the send|recv, but the effect was the same. Maybe I had to destroy dst/sync1/unsync2 to get syncoid to work.  It might not be a nice way to get back into sync, but deciding to sync it back was always going to require thought, and it is much better than resyncing the entire subtree. Once everything is renamed, syncoid doesn't know the difference.

As for the property name, the zfs-auto-snapshot script already uses com.sun:auto-snapshot property, and it has the advantage that it is already in use by people, and it not a double negative, and double negatives are almost always better avoided.

Maybe there should be separate properties for creating the actual snapshot, ie sanoid, and syncing the snapshot, ie syncoid.

I've currently using the com.sun:auto-snapshot property to generate an intermediate exclusion clauses for a generated sanoid.conf, so sanoid doesn't generate snapshots where the property is false.  I then use the property to cull the list of datasets for syncoid to replicate.  I'm just using some shell scripts to glue it together.

While it would be nice to integrate it, sanoid/syncoid are relatively simple, with a relatively simple configuration, and just work, and it may well be a good idea to keep them that way.  If you want to do down the path of properties, maybe znapzendz already does that well enough, though I haven't looked at it closely enough yet to comment on it.

AndrewJDR:
For my 2 cents, I'd really love to have something like this built-in to syncoid rather than needing to have bash glue or something similar to manage it.

phreaker0:
@attie for anyone who knows zfs this is obvious but if intermediate datasets are implemented it would be nice to have it documented.

properrty values are fine but I would drop the "some:" prefix, as it isn't needed. One would first check for "true", then "false" and everything else would be parsed as comma separated list.

@samarium nice workaround :-) I wouldn't reuse the "com.sun:auto-snapshot" property for syncing exclusion, at least for my use cases I would have  several datasets where a I don't wan't snapshots to be created regularly, but still want to be able to send them.
Furthermore I'm curious of that sanoid.conf generator, is it open source?

@AndrewJDR if you mean the "rename workaround", it would be possible to implement it but it wouldn't be trivial and it wouldn't be safe in case the operation is interrupted.

samarium:
@phreaker0 the point of the prefix is to segment the namespace so core zfs dataset properties won't ever clash with user properties, IIRC from reading many years ago in some SUN doco. Segmenting the namespace also means different apps/users don't clash either.

As for the sanoid.conf generator, it is just a quick hack that looks at a 3 way split of a sanoid.conf file, head, excl and tmpl, and uses com.sun.auto-snapshot=false and the head chunk to generate the excl sanoid conf chunk, then glues all the parts to a final sanoid.conf file.  It hasn't existing long enough to for me to soak test it, let along distribute it, and contains various names that are specific to my configuration, but you can take a look: https://gist.github.com/samarium/fd3b5470e7c1f2044d9953da5d7f18c7

AndrewJDR:
> For my 2 cents, I'd really love to have something like this built-in to syncoid rather than needing to have bash glue or something similar to manage it. This feels to me like a feature that is totally within syncoid's wheelhouse, but that's just my opinion.

@phreaker0 To be clear, I wasn't referring to the renaming thing in particular, I'm just saying that I very much hope a "syncoid:no-sync=" - like feature gets merged into syncoid proper, even if it has some limitations here and there.

I'd much prefer that over having to use a bash script to generate a config file, or similar.

samarium:
@AndrewJDR the config file stuff is to control if sanoid makes a snapshot in the first place, and can already be done manually by naming the dataset in the sanoid config file and tagging it with template = ignore.  I just automated it for my purposes using dataset properties.

What @attie has done is separate and doesn't need a config file, and you could merge his changes into your sanoid and use the modified syncoid now.

Even if the syntax changed at a later time, you can always update with something like:
zfs get -r property | awk/sed/perl/whatever-to-generate-zfs-set-commands | sh -x

AndrewJDR:
@samarium Ah ok, understood on the sanoid part. Regarding what @attie has done, I'm merely saying that I'd love to see it merged into syncoid proper at some point. I can of course maintain a separate fork of syncoid, but it's pretty nice not to have to do that.

phreaker0:
@samarium I didn't mean to drop the namespace prefix from the property name, but from the proposed value format. Thx for the generator.

@AndrewJDR yep, me too.

attie:
@samarium sounds good, thanks for the poke about double negatives... I was thinking that while writing it...

@phreaker0 thanks for your input... it sounds like people would appreciate it as it stands, and I'd certainly like the extra destination filter...

## todo:

1. Change `syncoid:no-sync` to `syncoid:sync`.
`true` is assumed if not set.

1. Implement filtering of remotes.
`syncoid:sync` value can be a comma separted list of remotes that **may** take a copy of this dataset.
If the remote isn't in the list, then it doesn't get the dataset

1. Keep contemplating handling of children whose parent isn't sync'd...

samarium:
If I replicate the properties as well as the data then the syncoid:sync property will replicate to the remote, which could cause issues if running syncoid on the remote.  As such I believe 
property name should be changeable via --property=name, that way we can use handle different sync policies on remote or indeed on a local sync.  This would also allow people who use a different software to generate snapshots using a different property as trigger to use the same property to trigger syncs with syncoid.

As for sync=false parent then sync=true children, it only matters when using -r, not with individual datasets.   If someone is going to be using sync=true/false and -r, then it is reasonable to expect people to be hands on, but it will get messier and messier with people wanting slightly different things.

I initially preferred to see sync=false cause a replication tree prune at that point, and anything under that is not automatically synced, but maybe if you really want to sync=true the children under the sync=false parent, then maybe you need a --sync-nosync-children option, or does the fact that you have had to manually create sync=true properties on each child prove sufficient intent, and as such justify creation of an empty remote dataset matching the sync=false parent to allow child replication?

Normally setting sync=false would cause it it be inherited by children, so normal case would sync=false, normal.  Only problem would be if replaced children from somewhere else, and replicated properties too, and sync=true was already set, but if you are up to this, then I expect you know what you are doing and that is what you want, and if you leave sync=true, then you get sync=true.

I supposed another option for children is sync=force, but not sure if it is worth the extra complexity.

attie:
I have inverted the logic, and implemented the true/false/${HOSTS} filtering as discussed above...

Children of no-sync datasets are still a concern.

attie:
Minimally tested script to migrate your properties... run at your own risk:

```
sudo zfs get -rt filesystem syncoid:no-sync ${POOL} -H -o name,value,source \
    | while read name value src; do \
        [ "${src}" != "local" ] && continue; \
        case "${value}" in \
            "true") value="false";; \
            "false") value="true";; \
            *) \
                echo "${name}: unhandled value '${value}'"; \
                continue;; \
        esac; \
        echo "${name} --> ${value}"; \
        sudo zfs set syncoid:sync="${value}" "${name}"; \
        sudo zfs inherit syncoid:no-sync "${name}"; \
    done
```

jimsalterjrs:
@phreaker0 you've been a lot more involved in this one than I have obviously - how do you feel about it in its current condition?

attie:
> it will skip syncing if the value is empty

@phreaker0 - could you expand?

I don't think it's possible to have an empty property, the following fail:

```bash
zfs set syncoid:sync tank/x
zfs set syncoid:sync= tank/x
```

Even if it did work, wouldn't this shift the issue to `getzfsvalue()` and parsing of the output from `zfs get`?

---

Admittedly the following works - but then we'd have a value of `""`, which isn't unset or `true` and is just weird/wrong - right?

```bash
zfs set syncoid:sync=\"\" tank/x
```

attie:
> Because of -s local, the syncoid:sync flag isn't being inherited

@AndrewJDR - I feel that this is counter-intuitive - properties are generally displayed as inherited, and using `-s local` will output _nothing_ if the property isn't explicitly set for that dataset... I've not tried, but suspect that would also cause issues in `getzfsvalue()` when trying to get `$values[2]`...

attie:
> datasets containing spaces won't work until this is rebased on master branch

Done

README.md
```diff
@@ -118,6 +118,30 @@ If ZFS supports resumeable send/receive streams on both the source and target th
...
+
+    This dataset will be synchronised to all hosts.
+
+  + `false`
+
+    This dataset will not be synchronised to any hosts - it will be skipped. This can be useful for preventing certain datasets from being transferred when recursively handling a tree.
+
+  + `host1,host2,...`
+
+    A comma seperated list of hosts. This dataset will only be synchronised by hosts listed in the property.
```
phreaker0: typo: seperated -> separated

redmop:
Zfs inherit can clear a property.

phreaker0:
@attie i just tried and for me 

`zfs set syncoid:sync= dataset`

works with ZFS on Linux @ master.

In the empty case the syncing will be skipped and the info message say's that it was skipped because the local hostname wasn't in the list. I would treat the empty value like "false" (with the corresponding info message), but this is solely personal preference.

@redmop 
but if the value is inherited again, one retrieves the inherited version and not the local cleared one.

README.md
attie: 👍 thanks

attie:
@redmop 
> Zfs inherit can clear a property.

Yes, but this return it to "_unset_", which outputs a dash `-`... correct?

attie:
@phreaker0 
> i just tried and for me
> `zfs set syncoid:sync= dataset`

Huh, thanks... I get:

```
$ sudo zfs set syncoid:sync= ell/test
missing value in property=value argument
usage:
        set <property=value> <filesystem|volume|snapshot> ...

[...]
$ echo $?
2
```

> In the empty case the syncing will be skipped and the info message say's that it was skipped because the local hostname wasn't in the list. I would treat the empty value like "false" (with the corresponding info message), but this is solely personal preference.

I agree...

jimsalterjrs:
Thanks everybody!

AndrewJDR:
> JDR - I feel that this is counter-intuitive - properties are generally displayed as inherited, and using `-s local` will output _nothing_ if the property isn't explicitly set for that dataset... I've not tried, but suspect that would also cause issues in `getzfsvalue()` when trying to get `$values[2]`...

@attie To rephrase: When using this property (e.g. `syncoid:sync=false`) on a container-only dataset where you still want the child datasets -- including any child datasets created in the future -- to be synced, it's pretty easy to forget to set `syncoid:sync=true` on any new child dataset that gets created in the future. This can lead to not replicating things that one wanted to be replicated.

I was merely advocating for two versions of the flag -- one which acts the same way as the current one, and the other which is checked in a non-recursive manner (`-s local`).

The two flags could be:
```
syncoid:sync
syncoid:sync-noinherit
```

With the syncoid:sync-noinherit flag using `-s local` for all checks. I'm not sure on the getzfs you mentioned... I haven't really had any issues with that on my patched version that uses `-s local`.

Anyway, does this not make sense? I'm curious why others wouldn't find this useful. Maybe I'm just using things in a weird way with regard to container-only datasets and replication...

AndrewJDR:
Hm, I'm still needing to use a local patch to add the functionality that I need for this, namely the `syncoid:sync-noinherit` that I mentioned above.

I still haven't heard back on my comments above... should I open a separate issue/PR for this?

PR #190, opened 2018-01-06
--quiet is not quiet

### LINKED ISSUE #132: syncoid --no-sync-snap --quiet not quiet
`syncoid --no-sync-snap --quiet` is not quiet and outputs something to stdout.

```
root@quartz:~# syncoid --no-sync-snap --quiet tank/home root@backup:tank/backup-quartz-tank-home
NEWEST SNAPSHOT: autosnap_2017-08-11_14:00:01_hourly
root@quartz:~# 
```

I expect that `--quiet` shouldn't output anything unless there is an error.

attie:
Should fix #132

I'm unsure about how you want to handle these prints in `monitor_health()` and `monitor_snapshots()`:
```
sanoid:93:      print "$message\n";
sanoid:173:     print "$msg\n";
```

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -597,7 +597,7 @@ sub getnewestsnapshot {
 	my $snaps = shift;
 	foreach my $snap ( sort { $snaps{'source'}{$b}{'creation'}<=>$snaps{'source'}{$a}{'creation'} } keys %{ $snaps{'source'} }) {
 		# return on first snap found - it's the newest
-		print "NEWEST SNAPSHOT: $snap\n";
+		if (!$quiet) { print "NEWEST SNAPSHOT: $snap\n"; }
 		return $snap;
 	}
 	# must not have had any snapshots on source - looks like we'd better create one!
--- a/syncoid
+++ b/syncoid
@@ -452,13 +452,13 @@ sub checkcommands {
 
 	if ($avail{'sourcecompress'} eq '') {
 		if ($compressargs{'rawcmd'} ne '') {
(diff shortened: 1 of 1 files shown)


syncoid
```diff
@@ -452,13 +452,13 @@ sub checkcommands {
 
 	if ($avail{'sourcecompress'} eq '') {
 		if ($compressargs{'rawcmd'} ne '') {
-			print "WARN: $compressargs{'rawcmd'} not available on source $s- sync will continue without compression.\n";
+			if (!$quiet) { print "WARN: $compressargs{'rawcmd'} not available on source $s- sync will continue without compression.\n"; }
```
martinvw: I personally would not surpres warnings, I would really love to see those in my mailbox

attie: Agreed - hence the separate commit - I wasn't sure how the maintainer(s) feel.

Going by the help however, this is probably technically correct:
```
  --quiet               Suppresses non-error output
```

---

I've seen the "_follow behaviour of other tools_" mentality on this project - anyone able/willing to compare with behaviour of `rsync`'s `--quiet` flag?
```
  -q, --quiet                 suppress non-error messages
```

jimsalterjrs:
It's OK to suppress warnings about `pv` or `mbuffer` not being available if `--quiet` is in effect - presumably, by that point, whoever's doing the system administration knows about it. I don't agree with suppressing warnings about a compression tool not being available, though - if you don't want it to gripe about the compression tool not being available, then set `--compression=none`!

attie:
Fair point, please see the update.

jimsalterjrs:
Thank you, folks!

Issue #105, opened 2017-07-12
Feature Request: Pre/Post scripting


redmop:
With pre/post scripting, I can quince filesystems of VMs, rsync something, or pauses databases before the snapshots.

redmop:
If the pre/post scripting are in the config sanoid.conf file, separate scripts can be made for each dataset, while using a wrapper around sanoid makes this very difficult as there isn't a way to "sanoid --dataset data/home" without making separate config directories.

jimsalterjrs:
I'd like more detail on this, please.  I'm sympathetic to the desire for it, but not sure how - for example - such a pre/post scripting option would handle recursion into child datasets.  Would it run once on the parent only, would it run for the parent and for each child, or would it need its _own_ recursion to be configurable separately from the overall recursion policy for the parent dataset?

I'm blocking on this until I get some workable pseudo-config-code from you, along with explanations of what it would do.  I don't need actual programmatic stuff, and you don't have to be perfect or pretty with names of directives or arguments, I'd just like to see a mocked up `sanoid.conf` that includes the kind of thing you'd like to do, with explanations of what your mocked-up `sanoid.conf` is doing and how.

I don't think _I'll_ ever use this, for any number of reasons, but it would probably be a fairly popular option, so if you can come up with a logical syntax for this, I'm willing to implement it.

redmop:
Using [BackupPC](http://backuppc.sourceforge.net/faq/BackupPC.html#_conf_dumppreusercmd_) as an example, why not do both?

sanoid.conf entry
```
[data/home]
     presnap=(run before any snapshots on this dataset)
     prechildsnap=(run before each child snapshot of that dataset)
     postchildsnap=(run after any snapshots on this dataset)
     postsnap=(run after any snapshots on this dataset)
```

Sanoid-run
```
process data/home
run presnap
run prechildsnap if recursive
take snapshot(s)
run postchildsnap if recursive
run postsnap
```

jimsalterjrs:
This looks doable.  Might want to ping me if you don't see something on this in a few days (or, take a stab at it yourself and give me a PR to review).

redmop:
I'll try to work something up this weekend.

dkruyt:
Give for the postchildsnap the volume as an parameter back to the script. This way each cild can run a script with it own options.

volumes for recursion

/zfspool/container/nameserver
/zfspool/container/mailserver
/zfspool/container/webserver

```
[/zfspool/containers]
[...]

PR #193, opened 2018-01-13
Add pre and post snapshot scripts

jouir:
Execute scripts before and after snapshot creation (#105).

Example with the following datasets:
```
# zfs list -o name -H
zfspool
zfspool/containers
zfspool/containers/dbserver
zfspool/containers/webserver
```
File `/etc/sanoid/sanoid.conf`:
```
[zfspool/containers]
	use_template = containers
	recursive = yes
	process_children_only = yes
	prescript = /opt/scripts/prescript.sh
	postscript = /opt/scripts/postscript.sh

[template_containers]
	autosnap = yes
	autoprune = yes
	hourly = 0
	daily = 30
	monthly = 0
	yearly = 0
```
Scripts:
```
# ls -l /opt/scripts/prescript.sh /opt/scripts/postscript.sh 
-rwxr--r-- 1 root root 113 Jan 13 12:56 /opt/scripts/postscript.sh
-rwxr--r-- 1 root root 113 Jan 13 12:56 /opt/scripts/prescript.sh
```
Script example:
```bash
#!/bin/bash
if [ $# != 1 ] ; then
    echo "Usage: $0 <snapshot>"
    exit 1
fi

SNAPSHOT=$1

# Do stuff
exit 0
```
Execution:
```
# ./sanoid --verbose --take-snapshots
INFO: taking snapshots...
executing prescript '/opt/scripts/prescript.sh zfspool/containers/dbserver@autosnap_2018-01-13_13:55:44_daily'
taking snapshot zfspool/containers/dbserver@autosnap_2018-01-13_13:55:44_daily
executing postscript '/opt/scripts/postscript.sh zfspool/containers/dbserver@autosnap_2018-01-13_13:55:44_daily'
executing prescript '/opt/scripts/prescript.sh zfspool/containers/webserver@autosnap_2018-01-13_13:55:44_daily'
taking snapshot zfspool/containers/webserver@autosnap_2018-01-13_13:55:44_daily
executing postscript '/opt/scripts/postscript.sh zfspool/containers/webserver@autosnap_2018-01-13_13:55:44_daily'
```
Snapshot created with pre and post scripts:
```
# zfs list -t all -o name -H
zfspool
zfspool/containers
zfspool/containers/dbserver
zfspool/containers/dbserver@autosnap_2018-01-13_13:55:44_daily
zfspool/containers/webserver
zfspool/containers/webserver@autosnap_2018-01-13_13:55:44_daily
```
It can be handy for some database services where service must be down before creating the snapshot to make it consistent.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -455,12 +455,35 @@ sub take_snapshots {
 
 	if ( (scalar(@newsnaps)) > 0) {
 		foreach my $snap ( @newsnaps ) {
+			my $dataset = (split '@', $snap)[0];
+			my $presnapshotfailure = 0;
+			if ($config{$dataset}{'pre_snapshot_script'} and !$args{'readonly'}) {
+				$ENV{'SANOID_TARGET'} = $dataset;
+				if ($args{'verbose'}) { print "executing pre_snapshot_script '".$config{$dataset}{'pre_snapshot_script'}."' on dataset '$dataset'\n"; }
+				if (system($config{$dataset}{'pre_snapshot_script'}) != 0) {
+					warn "WARN: pre_snapshot_script failed, $?";
+					$config{$dataset}{'no_inconsistent_snapshot'} and next;
+					$presnapshotfailure = 1;
+				}
+				delete $ENV{'SANOID_TARGET'};
+			}
 			if ($args{'verbose'}) { print "taking snapshot $snap\n"; }
 			if (!$args{'readonly'}) {
 				system($zfs, "snapshot", "$snap") == 0
 					or warn "CRITICAL ERROR: $zfs snapshot $snap failed, $?";
 				# make sure we don't end up with multiple snapshots with the same ctime
 				sleep 1;
+			}
+			if ($config{$dataset}{'post_snapshot_script'} and !$args{'readonly'}) {
+				if (!$presnapshotfailure or $config{$dataset}{'force_post_snapshot_script'}) {
+					$ENV{'SANOID_TARGET'} = $dataset;
+					if ($args{'verbose'}) { print "executing post_snapshot_script '".$config{$dataset}{'post_snapshot_script'}."' on dataset '$dataset'\n"; }
+					if (system($config{$dataset}{'post_snapshot_script'}) != 0) {
+						warn "WARN: post_snapshot_script failed, $?";
+						$config{$dataset}{'no_inconsistent_snapshot'} and next;
+					}
+					delete $ENV{'SANOID_TARGET'};
+				}
 			}
 		}
 		$forcecacheupdate = 1;
@@ -661,7 +684,7 @@ sub init {
 	tie my %ini, 'Config::IniFiles', ( -file => $conf_file ) or die "FATAL: cannot load $conf_file - please create a valid local config file before running sanoid!";
 
 	# we'll use these later to normalize potentially true and false values on any toggle keys
-	my @toggles = ('autosnap','autoprune','monitor_dont_warn','monitor_dont_crit','monitor','recursive','process_children_only');
+	my @toggles = ('autosnap','autoprune','monitor_dont_warn','monitor_dont_crit','monitor','recursive','process_children_only','no_inconsistent_snapshot','force_post_snapshot_script');
 	my @istrue=(1,"true","True","TRUE","yes","Yes","YES","on","On","ON");
 	my @isfalse=(0,"false","False","FALSE","no","No","NO","off","Off","OFF");
 
--- a/sanoid.conf
+++ b/sanoid.conf
@@ -67,6 +67,17 @@
 	daily_warn = 48
 	daily_crit = 60
 
+[template_scripts]
+	### run script before snapshot
+	### dataset name will be supplied as an environment variable $SANOID_TARGET
+	pre_snapshot_script = /path/to/script.sh
+	### run script after snapshot
+	### dataset name will be supplied as an environment variable $SANOID_TARGET
+	post_snapshot_script = /path/to/script.sh
+	### don't take an inconsistent snapshot
+	#no_inconsistent_snapshot = yes
+	### run post_snapshot_script when pre_snapshot_script is failing
+	#force_post_snapshot_script = yes
 
 [template_ignore]
 	autoprune = no
--- a/sanoid.defaults.conf
+++ b/sanoid.defaults.conf
@@ -15,6 +15,10 @@ path =
 recursive =
 use_template =
 process_children_only =
+pre_snapshot_script =
+post_snapshot_script =
+no_inconsistent_snapshot =
+force_post_snapshot_script =
 
 # If any snapshot type is set to 0, we will not take snapshots for it - and will immediately
 # prune any of those type snapshots already present.


jouir:
Hi, I'm glad you like my pull request!

I've rebased my branch and fixed conflicts. I've renamed prescript to pre-snapshot-script and postscript to post-snapshot-script. That makes sense to me.

For documentation, can you guide me where should I write it please? I've updated `sanoid.conf` with a template for scripts and `sanoid.defaults.conf` with empty defaults. I don't know where to put a detailed example.

jimsalterjrs:
OK, looking good so far! A couple more things before it goes into master, please. The first one is simple cosmetic only - please change the hash names to all lower case instead of the existing CamelCase, to match the style of the rest of the code.

The next one: please make sure your event handler handles recursion sanely. Part of this is going to mean setting an environment variable that the called scripts can read to tell which dataset or zvol is currently being handled. Perhaps you might set TARGET=path/to/dataset, which would allow the script to read $TARGET to figure out what it needs to quiesce or otherwise mess with. It should be possible to set pre-snapshot-script=/path/to/script in a recursively defined module and have the same script perform the same operations on multiple datasets/zvols.

Now some meatier ones: I don't like that you've configured it so that Sanoid refuses to take a snapshot at all if the pre-script fails. This is how Microsoft handles VSS; it refuses to take a volume shadow copy if a VSS writer fails to return, and in practice, this means a lot of Windows systems end up with no volume shadow copies at all. And it sucks! If this is a behavior you really like, you can have it, but it needs to be configurable and not a default. Default should be to always take snapshots. (It would probably make sense to refuse to run the *post*-snapshot-script if the pre-snapshot-script failed, though.) A sensible configure option name might be no-inconsistent-snapshots, with default value false. I'm open to discussion as to how this is implemented and what the configurable names are.

I'd also really, really like to see a configurable timeout value for these scripts. I don't like the idea that sanoid could get irretrievably hung from a script that didn't return, thereby stopping snapshots being taken at all for the entire system until the error is caught and fixed. This should be a fairly low value by default - maybe 5 seconds? I'm open to argument on the default value, but systems with 20+ (and even 100+) datasets are a definite target for sanoid to handle, so I would like to have a sane value here for those situations. I don't *think* this is something that can be handled with the system call itself; it's probably going to require a fork to implement. If Perl forking is something you're not familiar with and don't have the energy to learn on your own, talk to me, and I'll give you some pointers - it's a little confusing to wrap your head around at first but it's really not that hard once you understand it.

redmop:
A couple requests:

- Not running post-snapshot-script should pre-snapshot-script fail should be the default, but an option should exist to run the post-snapshot-script anyway.
- Script timeout: make that 5 seconds configurable.

jimsalterjrs:
> Not running post-snapshot-script should pre-snapshot-script fail should be the default, but an option should exist to run the post-snapshot-script anyway.

Works for me. What would you call it? `run-post-snapshot-script-on-fail`, or...? I'd like to keep these things human readable but not insanely verbose. Suggestions?

redmop:
That config name is pretty long. I don't know what I'd call it. You might need to include pre in there, and decide what happens if the snapshot itself fails, do we run the script or not?

jimsalterjrs:
Where are we with this one?

jouir:
> OK, looking good so far! A couple more things before it goes into master, please. The first one is simple cosmetic only - please change the hash names to all lower case instead of the existing CamelCase, to match the style of the rest of the code.

Speaking of coding style, I've made the following changes to my request:
* CamelCase is now converted to lowercase
* dashes (-) are converted to underscores (_) to match configuration options

> The next one: please make sure your event handler handles recursion sanely. Part of this is going to mean setting an environment variable that the called scripts can read to tell which dataset or zvol is currently being handled. Perhaps you might set TARGET=path/to/dataset, which would allow the script to read $TARGET to figure out what it needs to quiesce or otherwise mess with. It should be possible to set pre-snapshot-script=/path/to/script in a recursively defined module and have the same script perform the same operations on multiple datasets/zvols.

I was passing snapshot name as script argument but environment variables seem good too. Pre and post snapshot scripts can now look for `SANOID_TARGET` environment variable holding the dataset name (it was the snapshot name).

> Now some meatier ones: I don't like that you've configured it so that Sanoid refuses to take a snapshot at all if the pre-script fails. This is how Microsoft handles VSS; it refuses to take a volume shadow copy if a VSS writer fails to return, and in practice, this means a lot of Windows systems end up with no volume shadow copies at all. And it sucks! If this is a behavior you really like, you can have it, but it needs to be configurable and not a default. Default should be to always take snapshots. (It would probably make sense to refuse to run the post-snapshot-script if the pre-snapshot-script failed, though.) A sensible configure option name might be no-inconsistent-snapshots, with default value false. I'm open to discussion as to how this is implemented and what the configurable names are.

By default, Sanoid now takes eventually consistent snapshots unless `no_inconsistent_snapshot` configuration option is set. It will skip post snapshot script if pre script fails unless `force_post_snapshot_script` is set.

> I'd also really, really like to see a configurable timeout value for these scripts. I don't like the idea that sanoid could get irretrievably hung from a script that didn't return, thereby stopping snapshots being taken at all for the entire system until the error is caught and fixed. This should be a fairly low value by default - maybe 5 seconds? I'm open to argument on the default value, but systems with 20+ (and even 100+) datasets are a definite target for sanoid to handle, so I would like to have a sane value here for those situations. I don't think this is something that can be handled with the system call itself; it's probably going to require a fork to implement. If Perl forking is something you're not familiar with and don't have the energy to learn on your own, talk to me, and I'll give you some pointers - it's a little confusing to wrap your head around at first but it's really not that hard once you understand it.

I'm having troubles to implement this feature. I've tried with `alarm` trapped in `eval`, `fork`, `exec` and `system` functions but I'm not satisfied enough to submit it right now. I might work on this when free time will show up again. This is definitely a feature we need.

darkbasic:
Hi, what's the status of the patch? Does it apply/work on top of latest stable version and/or git master?
I want to use make 100% consistent hourly snapshots of VMs, without ever pausing them. So I exploit libvirt external snapshots backed by zfs to track the delta instead of qcow2. I save the ram dump on a secondary cheap drive. That way all of my VM backups are always 100% consistent and I'm also able to resume VMs on my backup server whenever I want to do maintenance on the primary machine.
But I *really* need pre-post hooks in order to do so.

darkbasic:
I've tried it, the branch works flawlessly. Unfortunately it doesn't expose the snapshot name, which is something I really need.
@jouir 	I opened a PR to expose the snapshot name through ENV variable: https://github.com/jouir/sanoid/pull/1

darkbasic:
I've added pruning hooks in my own fork: https://github.com/darkbasic/sanoid/commit/c9afce6830427aaa412d2dbe5190a90fedd0055a
I didn't open any more PRs because it looks like @jouir isn't interested anymore in pursuing mainline.

phreaker0:
@jouir are you still working on this or can i take a shot at the timeout logic?

jouir:
I recently had big changes in my personal life sucking up all of my time. I'm sorry for that. If you want to improve my work, feel free to do so. I will be glad to see this feature merged with or without me!

phreaker0:
@jouir np, thanks for the quick respone and your work on this feature.

darkbasic:
@jouir do you think you will be able to merge our PRs to your own fork (so can keep using this PR for upstream sanoid) or do you want me (or phreaker0) to open a new PR while cherry picking your past work?

jouir:
You can create your own PR. I won't be able to work on it soon.

PR #176, opened 2017-12-10
special char fixes

phreaker0:
This patches will escape all filesystem names, snapshot names and remote ssh commands as necessary to support all the characters which are allowed in ZFS to be used/supported for syncoid.
Furthermore several regular expression were adapted to ignore any meta characters in filesystem/snapshot names.

Fixes #40

PR #197, opened 2018-02-13
Resumable replication of interrupted zfs send/receive

phreaker0:
ZFS on Linux supports resume able zfs send/receive (since v0.7; and probably other zfs implementations too) in case the process is interrupted somehow (timeout, ...). This PR adds support for syncoid utilizing this awesome zfs feature if requested by the user via a comand line flag "--resume":

- for each dataset the target is checked for a receive_resume_token
- if one is found the zfs send/receive is resumed
- after this is finished successfully a ordinary sync is done, because the resumed zfs send will only be done to the next snapshot
- all the zfs receives will be instructed to save the state instead of discarding it in the case the resume flag is set (-s)
- if the feature is not supported by the target host, syncoid errors out

My primary use case is for initial and big incremental backup streams, i eventually want to switch my rsync based remote backups over to zfs send/receive. But for my private and also work uplinks it is almost sure the won't work through the whole initial zfs send/receive which probably take several weeks.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -919,7 +919,7 @@ sub getssh {
 	if ($fs =~ /\@/) {
 		$rhost = $fs;
 		$fs =~ s/^\S*\@\S*://;
-		$rhost =~ s/:$fs$//;
+		$rhost =~ s/:\Q$fs\E$//;
 		my $remoteuser = $rhost;
 		 $remoteuser =~ s/\@.*$//;
 		if ($remoteuser eq 'root') { $isroot = 1; } else { $isroot = 0; }
--- a/syncoid
+++ b/syncoid
@@ -97,7 +97,7 @@ if (! $args{'recursive'}) {
 	if ($debug) { print "DEBUG: recursive sync of $sourcefs.\n"; }
 	my @datasets = getchilddatasets($sourcehost, $sourcefs, $sourceisroot);
 	foreach my $dataset(@datasets) {
-		$dataset =~ s/$sourcefs//;
+		$dataset =~ s/\Q$sourcefs\E//;
 		chomp $dataset;
 		my $childsourcefs = $sourcefs . $dataset;
 		my $childtargetfs = $targetfs . $dataset;
@@ -126,11 +126,16 @@ exit 0;
 sub getchilddatasets {
 	my ($rhost,$fs,$isroot,%snaps) = @_;
 	my $mysudocmd;
+	my $fsescaped = escapeshellparam($fs);
 
 	if ($isroot) { $mysudocmd = ''; } else { $mysudocmd = $sudocmd; }
-	if ($rhost ne '') { $rhost = "$sshcmd $rhost"; }
+	if ($rhost ne '') {
+		$rhost = "$sshcmd $rhost";
+		# double escaping needed
+		$fsescaped = escapeshellparam($fsescaped);
+	}
 
-	my $getchildrencmd = "$rhost $mysudocmd $zfscmd list -o name -t filesystem,volume -Hr $fs |";
+	my $getchildrencmd = "$rhost $mysudocmd $zfscmd list -o name -t filesystem,volume -Hr $fsescaped |";
 	if ($debug) { print "DEBUG: getting list of child datasets on $fs using $getchildrencmd...\n"; }
 	open FH, $getchildrencmd;
 	my @children = <FH>;
@@ -142,6 +147,9 @@ sub getchilddatasets {
 sub syncdataset {
 
 	my ($sourcehost, $sourcefs, $targethost, $targetfs) = @_;
+
+	my $sourcefsescaped = escapeshellparam($sourcefs);
+	my $targetfsescaped = escapeshellparam($targetfs);
 
 	if ($debug) { print "DEBUG: syncing source $sourcefs to target $targetfs.\n"; }
 
@@ -209,8 +217,8 @@ sub syncdataset {
 		# if --no-stream is specified, our full needs to be the newest snapshot, not the oldest.
 		if (defined $args{'no-stream'}) { $oldestsnap = getnewestsnapshot(\%snaps); }
 
-		my $sendcmd = "$sourcesudocmd $zfscmd send $sourcefs\@$oldestsnap";
-		my $recvcmd = "$targetsudocmd $zfscmd receive -F $targetfs";
+		my $sendcmd = "$sourcesudocmd $zfscmd send $sourcefsescaped\@$oldestsnap";
+		my $recvcmd = "$targetsudocmd $zfscmd receive -F $targetfsescaped";
 
 		my $pvsize = getsendsize($sourcehost,"$sourcefs\@$oldestsnap",0,$sourceisroot);
 		my $disp_pvsize = readablebytes($pvsize);
@@ -245,7 +253,7 @@ sub syncdataset {
 			# $originaltargetreadonly = getzfsvalue($targethost,$targetfs,$targetisroot,'readonly');
 			# setzfsvalue($targethost,$targetfs,$targetisroot,'readonly','on');
 
-			$sendcmd = "$sourcesudocmd $zfscmd send $args{'streamarg'} $sourcefs\@$oldestsnap $sourcefs\@$newsyncsnap";
+			$sendcmd = "$sourcesudocmd $zfscmd send $args{'streamarg'} $sourcefsescaped\@$oldestsnap $sourcefsescaped\@$newsyncsnap";
 			$pvsize = getsendsize($sourcehost,"$sourcefs\@$oldestsnap","$sourcefs\@$newsyncsnap",$sourceisroot);
 			$disp_pvsize = readablebytes($pvsize);
 			if ($pvsize == 0) { $disp_pvsize = "UNKNOWN"; }
@@ -305,15 +313,15 @@ sub syncdataset {
 			# rollback target to matchingsnap
 			if ($debug) { print "DEBUG: rolling back target to $targetfs\@$matchingsnap...\n"; }
 			if ($targethost ne '') {
-				if ($debug) { print "$sshcmd $targethost $targetsudocmd $zfscmd rollback -R $targetfs\@$matchingsnap\n"; }
-				system ("$sshcmd $targethost $targetsudocmd $zfscmd rollback -R $targetfs\@$matchingsnap");
+				if ($debug) { print "$sshcmd $targethost $targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnap\n"; }
+				system ("$sshcmd $targethost " . escapeshellparam("$targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnap"));
 			} else {
-				if ($debug) { print "$targetsudocmd $zfscmd rollback -R $targetfs\@$matchingsnap\n"; }
-				system ("$targetsudocmd $zfscmd rollback -R $targetfs\@$matchingsnap");
+				if ($debug) { print "$targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnap\n"; }
+				system ("$targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnap");
 			}
 
-			my $sendcmd = "$sourcesudocmd $zfscmd send $args{'streamarg'} $sourcefs\@$matchingsnap $sourcefs\@$newsyncsnap";
-			my $recvcmd = "$targetsudocmd $zfscmd receive -F $targetfs";
+			my $sendcmd = "$sourcesudocmd $zfscmd send $args{'streamarg'} $sourcefsescaped\@$matchingsnap $sourcefsescaped\@$newsyncsnap";
+			my $recvcmd = "$targetsudocmd $zfscmd receive -F $targetfsescaped";
 			my $pvsize = getsendsize($sourcehost,"$sourcefs\@$matchingsnap","$sourcefs\@$newsyncsnap",$sourceisroot);
 			my $disp_pvsize = readablebytes($pvsize);
 			if ($pvsize == 0) { $disp_pvsize = "UNKNOWN"; }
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
Thanks for this, Christoff! Which distribution are you using? Ubuntu Xenial still doesn't have 7.0, and as far as I can tell there's no easy path to get there. I'd like to operationally test your work; will I need to spin up CentOS or similar VMs? Any ideas?

phreaker0:
@jimsalterjrs I'm using xenial on my desktop but I'm running zfs from master on this one. But on the servers I use Debian Stretch, by default they also still have v0.6.x but in their official stretch-backports repository they offer v0.7.x (currently v0.7.5) which I'm using.

jimsalterjrs:
Ah, thank you! I've never needed to build ZFS from master - were there any particular hiccups, or does it boil down to the usual **./configure ; make install** and Bob's your uncle?

Do you have breakages requiring manual intervention after, for example, a kernel upgrade?

phreaker0:
There were several minor hiccups, foremost with the bleeding edge features, but thats the point of running master of course but never complete disasters or lost data :) A couple of months ago their build system got support for generating debian packages with DKMS modules for spl/zfs so since than kernel upgrades mostly don't need manually recompiling the source again, which is very nice. But because I also use the latests stable kernel (which is actually the main reason why I'm running the latest version of zfs all the time on the desktop) you will still need to grab the latest master again which compiles fine again with the newer kernel.

jimsalterjrs:
Hrm. Honestly it sounds like I'd be better off spinning up a couple of 
CentOS VMs to test, since I don't really care about performance for this 
purpose. That sounds like a right PITA! =)

I really, really wish Canonical hadn't decided to make things so 
difficult to maintain a PPA. I feel like a second-class citizen stuck on 
0.6 after all this time.

gfriloux:
Hello!

I would love to see this feature added.
Any news ?

phreaker0:
@gfriloux 
The next Ubuntu LTS 18.04 is released on April 26th, which includes zfs v0.7.x, this may help testing the resume feature.

gfriloux:
Oh nice, this is pretty soon =)
Do you need tests to be done on FreeBSD ?

This feature is supposed to be there since FreeBSD 10.

phreaker0:
Yeah, tests on FreeBSD would be nice.

phreaker0:
This are the steps how I tested it:
- use two terminals
- terminal 1: syncoid --debug --resume storage/backup storage/backup_del
- if the sync progress is somewhere in the middle (pv progressbar):
- terminal 2: killall pv
- the zfs send command will now error out because of the broken pipe and will output a message that the data received is saved for a later resume
- terminal 1: syncoid --debug --resume storage/backup storage/backup_del
- the send is now resumed, but only to the next snapshot so after the resumed send is finished, another one is started (which again will save the state if interrupted)

jimsalterjrs:
OK. What happens if somebody tries to use --resume on a ZFS which does not support it on the source?
What happens if somebody tries to use --resume on a ZFS which does not support it on the target?

phreaker0:
Case1: assuming an non resumeable sync:
- source doesn't support resume: no error, received state will be saved on the target
- target doesn't support resume: error (because the receive command flag is unknown)
Case2: resuming an resumeable sync:
- source doesn't support resume: error (because the send command flag for providing the resume token is unknown)
- target doesn't support resume: not possible (receive token property is not set)

I didn't test it, but this should be the expected behaviour.

jimsalterjrs:
Ok. I'm going to queue up some more thorough testing in the next week. I 
want to make sure we handle older zfs versions non obnoxiously, and then 
I'd like to make resume default behavior, rather than requiring an argument.

Do you have much experience with resume in general? Does it seem solid on 
the ZFS end?

On April 23, 2018 15:52:36 Christoph Klaffl <notifications@github.com> wrote:
Case1: assuming an non resumeable sync:
source doesn't support resume: no error, received state will be saved on 
the target
target doesn't support resume: error (because the receive command flag is 
unknown)
Case2: resuming an resumeable sync:source doesn't support resume: error 
(because the send command flag for providing the resume token is unknown)
target doesn't support resume: not possible (receive token property is not set)
I didn't test it, but this should be the expected behaviour.

gfriloux:
Hello, i just tested here on FreeBSD 11.1-RELEASE, and it works!

When resuming: 
- I can see the script is calling `zfs send` with `-t`.
- I do get the "Resuming interrupted zfs send/receive from ..." and the remaining size to transfer seems good.

Ran test with a 20GB snapshot, that I killed multiple times (i did not kill `zfs` itself, but `pv`).

gfriloux:
As @jimsalterjrs said, it would be perfect if this feature could be autodetected on both ends, and use it whenever its possible.

gfriloux:
you can check if this feature is available by querying its property name : 
```
# zfs get receive_resume_token zroot/usr/home/user
NAME                       PROPERTY              VALUE      SOURCE
zroot/usr/home/user  receive_resume_token  -          -
```

It should return error code 0 on success, 2 if property doesnt exist.

phreaker0:
I just tested the cases with a debian stretch and debian jessie machine (0.7.6-1~bpo9+1 vs. 0.6.5.9-5~bpo8+1) and the result matches my expectations exactly.

@gfriloux thanks for testing!
@jimsalterjrs making it a default would be nice. So I will add a check for source/target support at start and an "--no-resume" flag to disable resume feature on user request, okay? I don't have used the resume much, so I unfortunately can't say how stable it is.

gfriloux:
To add little informations, i noticed resume as been backported to FreeBSD 10.3.

jimsalterjrs:
Checking for the resume token, then invoking by default unless the user specified --no-resume is perfect, thank you! If you'll make sure you're rebased against current, I'll pull the PR into mainstream this week once that's done.

phreaker0:
@jimsalterjrs Can you look at the other PR #176 and merge it before this one, because this is based on the former (special char escaping). I just checked and the branch is bases on your latest master. I will add and test the required changes.

phreaker0:
@jimsalterjrs 
latest push includes:
- made resumeable zfs send/receive a default
- updated documentation
- added zfs resume command check for source and target

jimsalterjrs:
Thank you! I'll start testing today. 

On 2018-04-25 09:13,
Christoph Klaffl wrote: 

> @jimsalterjrs [1]
> latest push includes: 
>

> * made resumeable zfs send/receive a default
> * updated
documentation
> * added zfs resume command check for source and target
>

> --
> You are receiving this because you were mentioned.
> Reply to
this email directly, view it on GitHub [2], or mute the thread [3].




Links:
------
[1] https://github.com/jimsalterjrs
[2]
https://github.com/jimsalterjrs/sanoid/pull/197#issuecomment-384281450
[3]

jimsalterjrs:
Hey @phreaker0 , can you look at my comments on #176 please? I went ahead and merged it because it doesn't break anything that isn't already broken, but there are some problems it _didn't_ fix that it should have, and it's probably easier for you to address it than me since you wrote it. =)

Specifically, it's not escaping the second source snapshot in a -I replication (which is fine when it's a syncoid-created snapshot, not fine when it's a user-generated snapshot because --no-sync-snap was invoked; also it should really just be escaped in all cases for consistency); also although you are calling escapeshellparam() for snapshot names in most cases, there's still an issue with generating the list, apparently, when there are spaces in the snapshot name.

Can you take a look at that before we go ahead with this resumable send/receive? Thank you!

jimsalterjrs:
OK @phreaker0, merged your fixes from #176. Can you resolve the conflicts now so I can test the resume function?

phreaker0:
@jimsalterjrs 
done

jimsalterjrs:
Can you fix the unintialized value warnings at 1092 and 1094 please?

```
root@bittybox:/usr/local/bin# syncoid-test --debug data/tmp data/tmp2
DEBUG: SSHCMD: /usr/bin/ssh    
DEBUG: checking availability of /usr/bin/lzop on source...
DEBUG: checking availability of /usr/bin/lzop on target...
DEBUG: checking availability of /usr/bin/lzop on local machine...
DEBUG: checking availability of /usr/bin/mbuffer on source...
DEBUG: checking availability of /usr/bin/mbuffer on target...
DEBUG: checking availability of /usr/bin/pv on local machine...
DEBUG: checking availability of zfs resume feature on source...
DEBUG: checking availability of zfs resume feature on target...
DEBUG: syncing source data/tmp to target data/tmp2.
DEBUG: checking to see if data/tmp2 on  is already in zfs receive using  /bin/ps -Ao args= ...
DEBUG: checking to see if target filesystem exists using "  /sbin/zfs get -H name 'data/tmp2' 2>&1 |"...
DEBUG: getting current value of receive_resume_token on data/tmp2...
  /sbin/zfs get -H receive_resume_token 'data/tmp2'
DEBUG: got receive resume token: 1-f66896133-d8-789c636064000310a500c4ec50360710e72765a526974005b0c9a7a515a782e41d989861f26c48f2499525a9c5407a8169823e36fd25f9e9a599290c0c7d3db39ca79c0e49f74092e704cbe725e6a63230a4249624ea97e416381457e625e767a6c4271625676496651697c41b19185ae81a98e81a5a5819985819195a191b42dc0a00490f21b4: 
Use of uninitialized value $par in substitution (s///) at /usr/local/bin/syncoid-test line 1092.
Use of uninitialized value $par in concatenation (.) or string at /usr/local/bin/syncoid-test line 1094.

```

jimsalterjrs:
Actually I got that one; you just don't test for empty @_ in escapeshellparam(). Don't worry about this, I'll fix it after I merge.

jimsalterjrs:
Found a more serious problem. When trying to do a pull replication from a remote host which does not support resume (although TBH, doesn't look like the resume has anything to do with it, I think this is going to happen with your special character fix regardless):

```
root@bittybox:/usr/local/bin# syncoid --debug root@192.168.100.157:target/dataset /data/tmp2
DEBUG: SSHCMD: /usr/bin/ssh    
DEBUG: checking availability of /usr/bin/lzop on source...
DEBUG: checking availability of /usr/bin/lzop on target...
DEBUG: checking availability of /usr/bin/lzop on local machine...
DEBUG: checking availability of /usr/bin/mbuffer on source...
DEBUG: checking availability of /usr/bin/mbuffer on target...
DEBUG: checking availability of /usr/bin/pv on local machine...
DEBUG: checking availability of zfs resume feature on source...
DEBUG: checking availability of zfs resume feature on target...
WARN: ZFS resume feature not available on source machine - sync will continue without resume support.
DEBUG: syncing source target/dataset to target /data/tmp2.
DEBUG: checking to see if /data/tmp2 on  is already in zfs receive using  /bin/ps -Ao args= ...
DEBUG: checking to see if target filesystem exists using "  /sbin/zfs get -H name '/data/tmp2' 2>&1 |"...
DEBUG: getting list of snapshots on target/dataset using /usr/bin/ssh     -S /tmp/syncoid-root-root@192.168.100.157-1524674477 root@192.168.100.157  /sbin/zfs get -Hpd 1 -t snapshot guid,creation ''"'"'target/dataset'"'"'' |...
DEBUG: target /data/tmp2 does not exist.  Finding oldest available snapshot on source target/dataset ...
DEBUG: getting estimated transfer size from source -S /tmp/syncoid-root-root@192.168.100.157-1524674477 root@192.168.100.157 using "/usr/bin/ssh     -S /tmp/syncoid-root-root@192.168.100.157-1524674477 root@192.168.100.157  /sbin/zfs send -nP ''"'"'target/dataset@1'"'"'' 2>&1 |"...
DEBUG: sendsize = 16127774720
INFO: Sending oldest full snapshot target/dataset@1 (~ 15.0 GB) to new target filesystem:
DEBUG: /usr/bin/ssh     -S /tmp/syncoid-root-root@192.168.100.157-1524674477 root@192.168.100.157 ' /sbin/zfs send '"'"'target/dataset'"'"'@'"'"'1'"'"' | /usr/bin/lzop  | /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null' | /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null | /usr/bin/lzop -dfc | /usr/bin/pv -s 16127774720 |  /sbin/zfs receive  -F '/data/tmp2'
```

jimsalterjrs:
Basically a ton of commands end up littered with both single and double quote encapsulation, which breaks the crap out of 'em. Can you fix please?

The fix I alluded to earlier for escapeshellparam() is just this:

```
sub escapeshellparam {
        my $par;
        if (scalar @_) {
            ($par) = @_;
            # "escape" all single quotes
            $par =~ s/'/'"'"'/g;
            # single-quote entire string
        }
        return "'$par'";
} 

```

phreaker0:
@jimsalterjrs what exactly does break or do you just mean we should skip quoting if there is no need for it? The long single and double quote encapsulation results from double escaping, for example:

- zfs list 'ssd/test/src'
- the above can be escaped reliable for remote ssh execution like:
- ssh remotehost 'zfs list '"'"'ssd/test/src'"'"'
- because in single quotes only the single quote symbol needs to be escaped which I do by ending the escape sequence, starting a new one with double quotes where I can safely use the single quote symbol, and then starting again the escaping of the rest with single quotes again
- I did it that way because it should take of all imaginable symbols, but of course for multiple escaping it won't look nice

jimsalterjrs:
@phreaker0  oh, I see - I mistakenly thought it was broken, because I had an error when I tried to use it.

Turns out that error was because I put a leading slash in my target name, d'oh!

jimsalterjrs:
This part's still a problem, but I'll clean it up myself now that I've merged your code:

```
Use of uninitialized value $par in substitution (s///) at ./syncoid line 1096.
Use of uninitialized value $par in concatenation (.) or string at ./syncoid line 1099.
```
It's what happens when you send an empty string to escapeshellparam(). It's just a WARN, but I have those turned on and I don't want to see them spamming up the console. :)

phreaker0:
thx, didn't think of a case where it would be empty

gfriloux:
Thx both of you for the work!

PR #201, opened 2018-02-20
implemented monitor-capacity flag for checking zpool capacity limits

phreaker0:
I just noticed that the monitor-health function doesn't check the capacity, I don't know why I assumed it does but I learnt the hard way that it didn't :-/ If you are okay to check the capacity too for determining the pool health it may be best to make the limit's configurable.
I hard coded some sane defaults for now:

- \>= 80%: Warning (the 80% rule of zfs)
- \>= 95%: Critical

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,11 @@ Which would be enough to tell sanoid to take and keep 36 hourly snapshots, 30 da
 
 	This option is designed to be run by a Nagios monitoring system. It reports on the health of the zpool your filesystems are on. It only monitors filesystems that are configured in the sanoid.conf file.
 
++ --monitor-capacity
+
+	This option is designed to be run by a Nagios monitoring system. It reports on the capacity of the zpool your filesystems are on. It only monitors pools that are configured in the sanoid.conf file. The default limits are 80% for the warning and 95% for the critical state. Those can be overridden by providing them
+	along like '=80,95". 
+
 + --force-update
 
 	This clears out sanoid's zfs snapshot listing cache. This is normally not needed.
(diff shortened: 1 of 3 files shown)


jimsalterjrs:
I'm definitely not going to be okay with hard-coded limits here.

For reference, Sanoid doesn't check this because standard Nagios plugins already can and do; I use the typical any-filesystem checks for capacity and usage.

phreaker0:
So you are okay with configurable capacity limits and which default to not checking, so nothing changes with a configuration change?

I use the standard nagios plugin too but I excluded zfs from it because I had problems on some hosts where it wouldn't work with an ordinary user because of some permission errors.

jimsalterjrs:
Yep, if configurable I'm down with it.

phreaker0:
@jimsalterjrs zpool capacity is now only checked if limits are provided along with the check flag like " --monitor-health=80,95". Is this acceptable to you?

phreaker0:
@jimsalterjrs yeah, will do that.

phreaker0:
@jimsalterjrs implemented the "--monitory-capacity" flag. Custom limits can be provided like:

`sanoid --monitor-capacity=90,96`

jimsalterjrs:
One last bit of donkeywork - Can they also be specified in the conf 
file? See defaults.conf for examples of configuration of warn and crit 
levels for monitor-snapshots.

phreaker0:
i will look into it

jimsalterjrs:
Thanks Christoph. As soon as we've got configurables available in the 
conf file and I can do a little testing, I'll push this into mainline.

I'm also planning to push your changes regarding when the cache file is 
regenerated; just got to find time to do some testing first.  I 
appreciate your work on these items!

phreaker0:
Limits will now be parsed from the configuration file (capacity_warn, capacity_crit). If not set the values from the default configuration file will be used as always.

Looking forward to getting things merged :-)

PR #203, opened 2018-02-23
cache improvements

### LINKED ISSUE #202: sanoid refreshing cache after each dataset when pruning ?
seems like sanoid prunes each dataset, then does a force rebuild of the cache?

With in the order of 500 datasets and 40000 snapshots, cache rebuild takes some time, especially where some pools are on HDD.

When I disable $forcecacheupdate, what was talking nearly an hour to complete, finishes in a couple of minutes.

```
diff --git a/sanoid b/sanoid
index d6e58ce..91d7f97 100755
--- a/sanoid
+++ b/sanoid
@@ -241,7 +241,8 @@ sub prune_snapshots {
                                                        }
                                                }
                                                removelock('sanoid_pruning');
-                                               $forcecacheupdate = 1;
+###TMP
+                                               $forcecacheupdate = 0;
                                                %snaps = getsnaps(%config,$cacheTTL,$forcecacheupdate);
                                        } else {
                                                print "INFO: deferring snapshot pruning - valid pruning lock held by other sanoid process.\n";
```
What is going to break if I leave this disabled?
Maybe I need to invoke sanoid --forcecacheupdate when I begin?
Am I going to need to do external locking to stop multiple sanoids running over each other?

phreaker0:
while pruning snapshots sanoid regenerates the cache file very often (depending on configuration size/dataset count/...). On pools with big number of snapshots and datasets this can be very time consuming (in my case with ~50000 snapshots and ~230 datasets a single regeneration takes ~50 seconds). This PR just removes the destroyed snapshots from the cache file instead of regenerating the whole thing. If another process holds the lock it will defer the update but finally (after all pruning is done) it will wait for the cache update lock and removes all the destroyed snapshots from the cachefile.
For one of my servers this lead to sanoid run time going down from 20-55 minutes to 1-2 minutes.

Fixes #202

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -42,6 +42,7 @@ my $forcecacheupdate = 0;
 my $cache = '/var/cache/sanoidsnapshots.txt';
 my $cacheTTL = 900; # 15 minutes
 my %snaps = getsnaps( \%config, $cacheTTL, $forcecacheupdate );
+my %pruned;
 
 my %snapsbytype = getsnapsbytype( \%config, \%snaps );
 
@@ -233,7 +234,6 @@ sub prune_snapshots {
 					# print "found some snaps to prune!\n"
 					if (checklock('sanoid_pruning')) {
 						writelock('sanoid_pruning');
-						my @pruned;
 						foreach my $snap( @prunesnaps ){
 							if ($args{'verbose'}) { print "INFO: pruning $snap ... \n"; }
 							if (iszfsbusy($path)) {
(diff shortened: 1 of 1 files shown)


phreaker0:
I'm running this on two server and my workstations and i don't see any issues yet, @samarium maybe you can test this too?

AndrewJDR:
I'm running this here and it's working great so far.
Thanks, it reduced my prune time dramatically.

PR #208, opened 2018-04-13
Added support for ZStandard compression.

danielewood:
Available in all major distros with a simple yum/apt-get/pkg.

References: 
ZSTD Compression by Allan Jude - https://www.youtube.com/watch?v=hWnWEitDPlM
Zstandard - https://facebook.github.io/zstd/


jimsalterjrs:
thank you!

PR #212, opened 2018-04-28
Restructure dist packages & bump rpm spec to 1.4.18

dcrdev:
-Move debian and rhel/fedora packaging into dedicated 'packages' directory.
-Bump rpm spec to 1.4.18

https://repo.dcrdev.com/el7/x86_64/sanoid-1.4.18-1.el7.centos.noarch.rpm
https://repo.dcrdev.com/el7/SRPMS/sanoid-1.4.18-1.el7.centos.src.rpm


DHowett:
(this may cause tools like `dpkg-buildpackage` to fail, as they _prefer_ debian/ to be in the root of the project. This can be worked around with `dpkg-buildpackage -R packages/debian/rules`)

Issue #213, opened 2018-04-29
Syncoid fails with remote shell tcsh/csh and mbuffer available


ecoutu:
I was using syncoid successfully targeting my rsync.net zpool successfully until I symlinked `ln -s /usr/local/bin/mbuffer /usb/bin/mbuffer`.

After that syncoid started to fail:

```
WARN: ZFS resume feature not available on target machine - sync will continue without resume support.
Sending incremental hyper/home/ecoutu/dev@autosnap_2018-04-29_12:00:01_hourly ... syncoid_deca_2018-04-29:13:28:27 (~ 1.0 MB):
1.48MiB 0:00:00 [25.3MiB/s] [=============================================================================================================================================================================================================================================] 144%            
Ambiguous output redirect.
CRITICAL ERROR:  /sbin/zfs send -I 'hyper/home/ecoutu/dev'@'autosnap_2018-04-29_12:00:01_hourly' 'hyper/home/ecoutu/dev'@'syncoid_deca_2018-04-29:13:28:27' | /usr/bin/pv -s 1073128 | /usr/bin/lzop  | /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null | /usr/bin/ssh     -S /tmp/syncoid-root-root@zfs.rsync.net-1525022904 root@zfs.rsync.net ' /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null | /usr/bin/lzop -dfc |  /sbin/zfs receive  -F '"'"'data1/home/ecoutu/dev'"'"'' failed: 256 at /opt/sanoid/syncoid line 380.
WARN: ZFS resume feature not available on target machine - sync will continue without resume support.
Sending incremental hyper/home/ecoutu/data@autosnap_2018-04-29_12:00:01_hourly ... syncoid_deca_2018-04-29:13:28:31 (~ 654 KB):
 577KiB 0:00:00 [10.3MiB/s] [================================================================================================================================================================================================================>                             ] 88%            
Ambiguous output redirect.
[...]

PR #214, opened 2018-04-29
Document compatibility with (t)csh shells

ecoutu:
Also, recommend installing mbuffer on FreeBSD systems when remote user
is using bourne compatible shell.

Documents jimsalterjrs/sanoid#213


ecoutu:
Just realized you said FREEBSD.readme in the ticket. Is it ok where it is instead?

ecoutu:
Updated the commit.

jimsalterjrs:
Thank you!

ecoutu:
The least I could do!

PR #216, opened 2018-04-30
Added sanoid:force-prune, syncoid:no-clone-rollback and syncoid:no-rollback

shodanshok:
I added some options to sanoid and syncoid:

- `force-prune` enable `sanoid` to skip the ZFS busy check when pruning snapshot. **Motivation:** when frequently running`syncoid` (ie: each minute), it is quite probable that a send/recv is in progress when  `sanoid` rotate snapshots. Actually, `sanoid` abort with a warning and refuse to prune any dataset snapshot. This patch enable `sanoid` to force pruning of non-synching (ie: old) dataset snapshot even if a send/recv is in progress;

- `no-clone-rollback` stops `syncoid` from rolling back clones on target host. **Motivation:** if clones exists, new data were created on the target host. Discarding them without warning seems a bad idea. When using the new `no-clone-rollback` option, the `zfs rollback` command is executed with `-r` (rollback snapshots) rather than `-R` (rollback both snapshots *and* clones). This, in turn, mean that the subsequent send/recv aborts with a warning stating that clones exist on the target dataset.

- `no-rollback` stops syncoid from rolling back both snapshots and clones on target host. It also disables forced receive ("-F"). **Motivation:** for hot-standby systems, where the backup host can take over the primary workload in case of disaster on first host, it is *very* important that under no circumstances the old primary box is permitted to rolling back the now-current backup host. Note that this option requires the target (standby) host to be in readonly mode or at least with `atime` disabled (otherwise, a simple `ls` on the backup host will update dir/file access time, preventing further synchronizations from the primary host).


redmop:
Please update the readme as well.

syncoid
```diff
@@ -359,13 +359,17 @@ sub syncdataset {
...
 			# rollback target to matchingsnap
+			my $rollbacktype="-R";
+			if (defined $args{'no-clone-rollback'}) {
+				$rollbacktype = "-r";
+			}
 			if ($debug) { print "DEBUG: rolling back target to $targetfs\@$matchingsnap...\n"; }
 			if ($targethost ne '') {
-				if ($debug) { print "$sshcmd $targethost $targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnapescaped\n"; }
-				system ("$sshcmd $targethost " . escapeshellparam("$targetsudocmd $zfscmd rollback -R $targetfsescaped\@$matchingsnapescaped"));
+				if ($debug) { print "$sshcmd $targethost $targetsudocmd $zfscmd rollback $rollbacktype $targetfs\@$matchingsnap\n"; }
```
phreaker0: you dropped the escaped cmd, fs and snapshot names. You need to add them back or else fs and snapshots with specials characters won't work anymore.

shodanshok:
@phreaker0 You are right, I was working on a somewhat older file. I fixed it, thanks!

@redmop I've updated the README file, thanks.

shodanshok:
Hi, anyone with some comments/suggestions about the patches above?
Thanks.

sanoid
```diff
@@ -1082,6 +1082,7 @@ Options:
   --monitor-snapshots   Reports on snapshot "health", in a Nagios compatible format
   --take-snapshots      Creates snapshots as specified in sanoid.conf
   --prune-snapshots     Purges expired snapshots as specified in sanoid.conf
+  --force-prune		Purges expired snapshots even if a send/recv is in progress
```
phreaker0: indentation should be in line with the others

syncoid
```diff
@@ -1139,6 +1152,8 @@ Options:
   --target-bwlimit=<limit k|m|g|t>  Bandwidth limit on the target transfer
   --no-stream           Replicates using newest snapshot instead of intermediates
   --no-sync-snap        Does not create new snapshot, only transfers existing
+  --no-clone-rollback	Does not rollback clones on target
```
phreaker0: indentation should be in line with the others

phreaker0:
@shodanshok 

force-prune: Does it work to remove snapshots while they are sent? I guess because the snapshot deleting is done in the background asynchronously in newer zfs implementations it should work and the deleting is delayed until the send is finished. Did you try that?

it may be based to rebase this PR to the latest master to get rid of some confusing commits

shodanshok:
@phreaker0 Hi, thanks for your comments! I rebased my patches against the lastest `jimsalterjrs:master`
On latest code, I can not find the indentation errors you noticed above. Am I missing something?

Regarding `force-prune`: the original code does not try at all to delete *any* snapshot if a zfs send/recv is in progress for the dataset (check the `iszfsbusy` function). In other words, it aborts even if destroying an old snapshot is a safe operation. My code adds an option to skip this check, relying on ZFS to actually identify (and rejecting) any deletion which will interfere with the current send/recv. Any `zfs destroy <snap>` against a snapshot which is in the sending chain will be immediately aborted by ZFS itself with a message like `cannot destroy snapshot <name>: dataset is busy`. In other words, async destroy seems to be about delaying the actual I/O operations, rather than delaying the logica operation (ie: `zfs destroy`) itself.

Back to the PR: should I close this PR and open a new one against current master code?

Thanks.

phreaker0:
I see the over indentation and under indentation in the github diff view, the comments are right after the affected lines.

I actually didn't mean to rebase but squash the commits, but you don't need to do that, as I just read that github supports this upon merge, so @jimsalterjrs can decide.

Okay now understand better, @jimsalterjrs  is there any reason to not make it default to rely on the zfs commands to deny the pruning by default?

shodanshok:
@phreaker0 @jimsalterjrs I just realized another case where `no-rollback` would be very useful: to prevent accidental source rollbacks to affect backup on destination host. For example, give following the source and destination dataset:

```
[root@localhost sanoid]# zfs list -t all
NAME                                                          USED  AVAIL  REFER  MOUNTPOINT
tank                                                          294K  1.86G  25.5K  /tank
tank/test                                                      24K  1.86G    24K  /tank/test
tank/test@snap1                                                 0B      -    24K  -
tank/test@snap2                                                 0B      -    24K  -
tank/test@snap3                                                 0B      -    24K  -
tank/test@snap4                                                 0B      -    24K  -
tank/test@syncoid_localhost.localdomain_2018-06-29:00:14:12     0B      -    24K  -
```

rolling back the source to snap1 via `zfs rollback tank/test@snap1 -r` and sending it using default sanoid parameter (ie: without `--no-rollback`) cause the destination to also rolling back to snap1:

```
[root@localhost sanoid]# ./syncoid tank/test root@192.168.0.110:tank/test --debug
...
DEBUG: rolling back target to tank/test@snap1...
/usr/bin/ssh     -S /tmp/syncoid-root-root@192.168.0.110-1530224215 root@192.168.0.110  /sbin/zfs rollback -R 'tank/test'@'snap1'
...
```

Indeed, destination now look as the following:

```
[root@localhost ~]# zfs list -t all
NAME                                                          USED  AVAIL  REFER  MOUNTPOINT
tank                                                         1.12M  1.86G  25.5K  /tank
tank/test                                                      24K  1.86G    24K  /tank/test
tank/test@snap1                                                 0B      -    24K  -
tank/test@syncoid_localhost.localdomain_2018-06-29:00:16:56     0B      -    24K  -
[root@localhost ~]# zfs list -t all
NAME                                                          USED  AVAIL  REFER  MOUNTPOINT
tank                                                         1.12M  1.86G  25.5K  /tank
tank/test                                                      24K  1.86G    24K  /tank/test
tank/test@snap1                                                 0B      -    24K  -
tank/test@syncoid_localhost.localdomain_2018-06-29:00:16:56     0B      -    24K  -

```

The fact that modifying the source has a direct influence on backups stored on destination dataset is quite concerning. Using `--no-rollaback` no harm will be done, at the cost of more "fragile" replication (ie: any modification on destination host will abort the replication with a warning).

jimsalterjrs:
> is there any reason to not make it default to rely on the zfs commands to deny the pruning by default?

I'm actually not sure I understand what's being asked for here.

The original code is belt-and-suspenders-and-training-wheels safe, by design. It will not delete ANY snapshot, under any circumstances, until after a replication has completed successfully. This makes sure that it won't get rid of anything that might be necessary to keep source and destination capable of incremental replication. Once a replication _has_ successfully completed, it gets rid of all `syncoid_hostname` snapshots but the most recent one, which will now be the common snapshot future incrementals can be based on from the same hostname.

Keep in mind that getting rid of `@syncoid` snapshots in any other way is *dangerous*, even if a partial replication has succeeded - a partial replication might move only a couple hourlies that will be gone in a day; if the `@syncoid` snapshot is removed without a new common `@syncoid` snapshot being present on both ends, you can then end up with a broken replication schema that will have to be replaced with a costly full to rebase.

shodanshok:
@jimsalterjrs The problem I'm fixing with `--force-prune` is related to how `sanoid` and `syncoid` interact.

Scenario: a dataset is being synchronized to a remote server using `syncoid` at very small time interval (ie: 30 seconds). At the same time, the source server runs `sanoid` to regularly take and rotate snapshots. The problem is that, when `sanoid` needs to rotate (ie: delete) an old snapshot (ie: a very old tank/test@autosnap), it is very problable that, due the very frequent `syncoid` invocations, the same dataset (tank/test) is being replicated on the remote side. The old source snapshot is *not* required (it was transferred may days ago, maybe), but `sanoid` aborts with a `send receive is in progress` warning. Basically, the `iszfsbusy()` checks for any send/recv for the *entire dataset*, while we should only care for snapshots which are (in the process to be) in transfer.

`--force-prune` skips the send/recv test when deciding if an old snapshot can be deleted, leaving the "final decision" to zfs itself: if the *snapshot*, rather than the dataset, is required for the correct completition of a send/recv operation, it will abort the deletion with a `dataset is busy` warning.

I have in production `--force-prune` since some months and I can not see any problem with it.

shodanshok:
@jimsalterjrs Any changes to review the patches? Thanks.

shodanshok:
@phreaker0 @jimsalterjrs any update/opinion on this PR?

sanoid
```diff
@@ -293,7 +293,7 @@ sub prune_snapshots {
 						writelock('sanoid_pruning');
 						foreach my $snap( @prunesnaps ){
 							if ($args{'verbose'}) { print "INFO: pruning $snap ... \n"; }
-							if (iszfsbusy($path)) {
+							if (iszfsbusy($path) && !$args{'force-prune'}) {
```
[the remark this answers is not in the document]
shodanshok: Sure.

shodanshok:
@phreaker0 I have applied the suggested change.

phreaker0:
@shodanshok thx, can you also fix the two mentioned indentation deviants from above (in the help text)

phreaker0:
thx
@jimsalterjrs I think it would be best to squash the commits on this one upon merge to keep the history clear

jimsalterjrs:
@phreaker0 moar rebase

shodanshok:
@phreaker0 I just did it; can you check now? And a big thank for taking the time to review my patches.

syncoid
```diff
@@ -20,7 +20,7 @@ my %args = ('sshkey' => '', 'sshport' => '', 'sshcipher' => '', 'sshoption' => [
 GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r",
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
-                   "no-clone-handling", "no-privilege-elevation", "force-delete") or pod2usage(2);
+                   "no-clone-handling", "no-privilege-elevation", "force-delete" "no-clone-rollback", "no-rollback") or pod2usage(2);
```
phreaker0: syntax error, missing ','

shodanshok:
@phreaker0 Done, thanks.

PR #217, opened 2018-04-30
Do not require user to be specified for syncoid

aerusso:
1. This brings the behavior of syncoid into line with the documentation, which suggests that `:`, rather than `@`, is required when specifying a remote dataset. The behavior of a TARGET or SOURCE that begins with `:` is altered to allow for local dataset names with colons in them. An example is provided in the documentation.

2. The name of the control socket is changed slightly.

3. We lose the ability to always determine if the ssh-ed user is root. This is still backwards compatible (because those configurations were not supported before), and furthermore will still function fine so long as `sudo` is installed. (I also have another PR lined up to add an option to disable `sudo`).

## Testing
Not a whole lot. It superficially runs on my configurations, and I also checked that the colon-at-the-start works in a couple cases.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -930,20 +930,22 @@ sub getssh {
 	my $isroot;
 	my $socket;
 
-	# if we got passed something with an @ in it, we assume it's an ssh connection, eg root@myotherbox
-	if ($fs =~ /\@/) {
+	# if we got passed something with an : in it, but not at the beginning,
+	# we assume it's an ssh connection, eg root@myotherbox:pool/dataset
+	if ($fs =~ /^[^:]+:/) {
 		$rhost = $fs;
-		$fs =~ s/^\S*\@\S*://;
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
I don't really want to take this one right now, sorry. For the moment I'm comfortable with requiring a user to be specified when using an ssh tunnel - particularly given that the necessary user will so very frequently NOT be the currently-logged-in user, so I don't want to do SSH's typical "just try the current username if none specified" behavior.

aerusso:
@jimsalterjrs Hey, thanks for looking at it and giving your reasoning. I just wanted to make sure I understand it. Ideally, I would think you would want to specify the behavior of each target host once (presumably in an ssh configuration file), rather than in every call that uses ssh. 

Also, this was going to be the first of a sequence of patches that added support for recent support for user-level permissions (which I've now been using for a month).

PR #233, opened 2018-06-19
Skip parent

phreaker0:
New option "--skip-parent" for skipping the parent dataset from recursive replication.

Fixes #231

PR #235, opened 2018-06-19
added option for using extra identification in the snapshot name for …

### LINKED ISSUE #225: Support for multiple targets
I want to replicate my zpool to three different targets (usb drives).
That means I need to keep three sync states for syncoid to be able to sync against a drive that has not been connected for the last couple of sync runs.
I can see, that syncoid creates snapshots with the name syncoid-<hostname>_<datetime>.
Is it possible to use another label (like disk1, disk2 and disk3 for example) instead of the hostname or append a suffix to the snapshot label?

phreaker0:
…replicating to multiple targets.

Fixes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 + --identifier=
 
-	Adds the given identifier to the snapshot name after "syncoid_" prefix and before the hostname. This enables the use case of reliable replication to multiple targets from the same host. The following chars are allowed: a-z, A-Z, 0-9 and -.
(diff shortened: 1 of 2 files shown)


jimsalterjrs:
Hey Christoph, conflicts with some of your other patches which I've accepted. Resolve please?

phreaker0:
I will wait for merged #233 , as it would conflict again

phreaker0:
I resolved the conflicts and allowed all characters which are supported by zfs snapshots to be used.

jimsalterjrs:
Looks good from here. I assume you've tested?

phreaker0:
Yes, I'm actually using it in production already for remote backup stuff.

PR #243, opened 2018-07-17
Fix 'resume support' detection on FreeBSD

pit3k:
FreeBSD does not accept options at the end, so without this fix it always fails:

```
$ uname -sr
FreeBSD 11.2-RELEASE
$ /sbin/zfs get receive_resume_token -d 0
cannot open '-d': dataset does not exist
cannot open '0': dataset does not exist
$ /sbin/zfs get -d 0 receive_resume_token
NAME      PROPERTY              VALUE      SOURCE
newzroot  receive_resume_token  -          -
```

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -614,7 +614,7 @@ sub checkcommands {
 
 	# check for ZFS resume feature support
 	if ($resume) {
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
Thank you for the patch, @pit3k. Merged.

PR #247, opened 2018-07-27
implemented support for zfs bookmarks as fallback

### LINKED ISSUE #88: feature request for syncoid: use bookmark feature as fallback method
I hope [this reddit discussion](https://www.reddit.com/r/zfs/comments/5op68q/can_anyone_here_explain_zfs_bookmarks/) makes it clear that the main advantage to `bookmark` in relation with `zfs send` is that you can afford to lose the source snapshot yet still be able to bring the destination in sync with the source.

It would be very useful if you can add the ability use bookmarks as a fallback mechanism:
when syncoid finds a common snapshot between source and destination, it should automatically bookmark the matching source snapshot.
if syncoid does not find a common snapshot, try to find a bookmark created by syncoid and that fails too, try to find newest bookmark and attempt to send incrementally from that bookmark to latest snapshot.
if that fails, try again with the second newest bookmark it finds, and repeat until oldest bookmark. only then fail if the incremental from oldest bookmark to latest snapshot still fails.

beware that when sending bookmarks you cannot have `-v` and [cannot estimate the size](https://utcc.utoronto.ca/~cks/space/blog/solaris/ZFSBookmarksMechanism) of the sending stream

also pay attention to the fact that -I and -R [do not work with sending bookmarks](https://web.archive.org/web/20160618005021/http://comments.gmane.org:80/gmane.comp.file-systems.openzfs.devel/514), only -i

all in all the use of bookmarks may seem to complicate things but in the end you will have more happy users by not having to re-send datasets after accidental snapshot deletions. and syncoid will be the first serious replication tool that fully takes advantage of the bookmarks feature :) (I only know about [zap](https://github.com/Jehops/zap) but it seems kind of immature)

phreaker0:
If no matching snapshots are found for replication, syncoid will now check for matching bookmarks on source with snapshots on the target (oldest first). If found there are two cases:
* no-stream: direct replication to latest snapshot
* stream: replication to next successor of bookmarked snapshot, after that ordinary replication with intermediate snapshots as bookmark sync doesn't support intermediates

I didn't test yet what happens on system where zfs bookmarks are not supported but it should just silently ignore that and show the ordinary sync error. In the case only the target doesn't support bookmarks, there will be a sync error with the bookmarks.

I also added two tests which I used for testing/developing this feature. They can be safely ran without the need of a vm (no time mangling, only zpool creation with temporary files).

Fixes #88

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -370,11 +370,48 @@ sub syncdataset {
 
 		my $targetsize = getzfsvalue($targethost,$targetfs,$targetisroot,'-p used');
 
-		my $matchingsnap = getmatchingsnapshot($sourcefs, $targetfs, $targetsize, \%snaps);
+		my $bookmark = 0;
+		my $bookmarkcreation = 0;
+
+		my $matchingsnap = getmatchingsnapshot($sourcefs, $targetfs, \%snaps);
 		if (! $matchingsnap) {
-			# no matching snapshot; we whined piteously already, but let's go ahead and return false
-			# now in case more child datasets need replication.
-			return 0;
+			# no matching snapshots, check for bookmarks as fallback
+			my %bookmarks = getbookmarks($sourcehost,$sourcefs,$sourceisroot);
+
+			# check for matching guid of source bookmark and target snapshot (oldest first)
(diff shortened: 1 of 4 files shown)


jimsalterjrs:
I'm seeing conflicts here. Am I missing a dependency PR, or does this need to be rebased vs current?

PR #248, opened 2018-07-29
force delete flag

### LINKED ISSUE #86: feature request: `--force-delete` option to "cowardly destroy destination" if no common snapshot is found
````
CRITICAL ERROR: Target exists but has no matching snapshots!
                Replication to target would require destroying existing
                target. Cowardly refusing to destroy your existing target.
````

^ please, for the above case, add a `--force-delete` command line option.
because in the end I would manually have to delete the destination anyway :)

phreaker0:
"--force-delete" flag will instruct syncoid to destroy target datasets which don't have matching snapshots/bookmarks so replication works.

Note: this is bases on another PR because it shares some common code, so the former should be merge first

Requires/Based on PR: #247
Fixes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,10 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 	This argument tells syncoid to not use resumeable zfs send/receive streams.
 
++ --force-delete
+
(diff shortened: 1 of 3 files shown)


phreaker0:
I added documentation and error checking for snapshot/bookmark listing errors, as the later is really dangerous with "--force-delete" if not captured.

jimsalterjrs:
@phreaker0 moar rebase

phreaker0:
done

PR #255, opened 2018-08-23
Add option --no-mount

LordAro:
Makes use of -u with zfs receive so that datasets on the other end are not automatically mounted


LordAro:
Noticed that I didn't add it to the helptext - added and rebased

@jimsalterjrs I understand you're busy, perhaps in the interests of development speed someone like @phreaker0 could be given merge rights?

jimsalterjrs:
needs rebasing

TerraTech:
Would seem that #295 will provide for this ability therefore the new switch '--no-mount' is no longer necessary.  FWIW, I almost started doing the same for '-u' and '-o property=value', '-x property' and then I saw the referenced pull request which is all encompassing.

phreaker0:
@LordAro It's now possible with the merged functionality in PR #295 and #343. I hope this works for you.

PR #254, opened 2018-08-15
Add xz compression

spheenik:
I needed a way to compress as much as possible, since I want to push a snapshot over a slow link (1MB/sec upload). So I added xz compression. I tested it here already, works pretty well.

PR #263, opened 2018-09-05
prepare for v1.4.19 release

phreaker0:
As requested in #254


phreaker0:
@jimsalterjrs ping :-)

jimsalterjrs:
@phreaker0 I'm sorry, I'm probably making your life harder approving things out-of-order. If we need what's in here, it'll need to get re-based... but you may want to hang on doing it until I work my way through the rest of the PRs that merge cleanly. I'll ping you again in a bit.

phreaker0:
will do! No problem, I'm excited that this moves forward!

jimsalterjrs:
@phreaker0 I've merged everything that I could merge without rebasing, now would be a good time to rebase the PRs we want to accept. Can you work on that? I want to get the PR list cleaned up this afternoon, then do some testing so we can push out a whopper of a release with all this great work I know you and a few other folks have been using out-of-tree for a while now.

phreaker0:
@jimsalterjrs I just added all the changes from the recent merges, i didn't notice that you already added the recent onse to the generic changelist. Should I use your descriptions for the recent changes to resolve the conflict?

jimsalterjrs:
I like your more detailed stuff honestly. Biggest problem is the version number though; there's a reason I hadn't assigned one yet.

This version can't be 1.4.19; by semantic versioning we need a 2.0.0 due to the backwards-incompatible required update on sanoid.defaults.conf IMO. Thoughts? I might be overthinking that and a 1.5.0 is sufficient, but I know this definitely amounts to much more than "patches". =)

https://semver.org/

jimsalterjrs:
Fair. Let's plan on 2.0.0 then, and unless I'm missing something, you covered everything in your proposed CHANGELIST that I did, but in more detail giving credit to authors, so let's go with your version (but with the updated version number).

We'll need to fix the version number in sanoid and syncoid themselves to match, of course.

Are you ready to feature-freeze for this version number?

phreaker0:
@jimsalterjrs yeah, feature freeze sounds good. #300 should be merged to prevent the introduced warnings. I will update this PR.

PR #264, opened 2018-09-05
ignore unknown interval type to prevent perl warnings

phreaker0:
In my case, I have "frequently" key values in the config, but the release version doesn't support them but they are iterated and used which leads to perl warnings about uninitialized values. So just skip processing unknown interval types.


jimsalterjrs:
I don't _want_ to approve this; it makes my teeth itch. Do you still need this, with the more-frequent-interval PR approved and merged? I have a pretty strong feeling that config parameters that Sanoid doesn't understand should produce a warning, preferably a human-readable one. Thoughts?

phreaker0:
@jimsalterjrs 
Yeah a warning is fine, i added it. I just don't wanted sanoid to try to handle an unknown interval type which produced perl errors.

PR #267, opened 2018-09-05
parse values for recursive key like the others booleans, previously a…

### LINKED ISSUE #219: "recursive = false" in sanoid.conf taken as true
It seems that any non-empty value given to `recursive` is taken as a true value.

Also related, according to [the TOML docs](https://github.com/toml-lang/toml#user-content-boolean), you're supposed to be using `true` in Boolean values, not `yes`. I'm filing it here instead of as a separate bug because I suspect it's a symptom of the same problem, not handling TOML Booleans correctly.

phreaker0:
…ny non empty value was considered as true

Fixes #219

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -743,7 +743,7 @@ sub init {
 
 		# how 'bout some recursion? =)
 		my @datasets;
-		if ($ini{$section}{'recursive'}) {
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
this one will need rebasing.

phreaker0:
done

PR #269, opened 2018-09-11
Ignore snapshots types that are set to 0.

muff1nman:
I am only using monthly snapshots and would like to use `--monitor-snapshots` to ensure operation. Below shows my config:

```
[tank/Backup]
  use_template = backup

#############################
# templates below this line #
#############################

[template_backup]
  hourly = 0
  daily = 0
  monthly = 6
  yearly = 0
  autosnap = yes
  autoprune = yes

[template_replica]
  hourly = 0
  daily = 0
  monthly = 8
  yearly = 0
  autosnap = no
  autoprune = yes
```

However, `--monitor-snapshots` fails with:

```
CRIT: tank/Backup has no daily snapshots at all!, CRIT: tank/Backup has no hourly snapshots at all!
```

In this case, I would expect to not have daily or hourly snapshots. This PR fixes the above by skipping checks for snapshot types where the number of snapshots is 0 (or negative - I wasn't sure if that was a valid use case).

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -127,6 +127,7 @@ sub monitor_snapshots {
 
 		my @types = ('yearly','monthly','daily','hourly');
 		foreach my $type (@types) {
+			if ($config{$section}{$type} == 0) { next; }
 
 			my $smallerperiod = 0;
 			# we need to set the period length in seconds first


redmop:
Maybe just do "0" here. -1 can sometimes mean keep everything. If that change happens here, it will make monitoring very unhappy.

phreaker0:
@muff1nman there are extra configuration options for monitoring thresholds, in your case you would need something like:

```
hourly_warn = 0
hourly_crit = 0
daily_warn = 0
daily_crit = 0
monthly_warn = 32
monthly_crit = 35
yearly_warn = 0
yearly_crit = 0
```

Regarding your PR, it also makes sense as I can't figure out a use case where one wants monitoring of snapshots which are not taken by sanoid itself.

muff1nman:
@phreaker0 can you explain a bit more? I am doing exactly that with the following configurations and using this PR:

Server A has the following sanoid.conf:

```
[tank/Backup]
  use_template = backup

#############################
# templates below this line #
#############################

[template_backup]
  hourly = 0
  daily = 0
  monthly = 6
  yearly = 0
  autosnap = yes
  autoprune = yes

[template_replica]
  hourly = 0
  daily = 0
  monthly = 8
  yearly = 0
  autosnap = no
  autoprune = yes
```

Server B has the same except the name of the replica pool is different:

```
[backroom/Backup]
  use_template = replica

#############################
# templates below this line #
#############################

[template_backup]
  hourly = 0
  daily = 0
  monthly = 6
  yearly = 0
  autosnap = yes
  autoprune = yes

[template_replica]
  hourly = 0
  daily = 0
  monthly = 8
  yearly = 0
  autosnap = no
  autoprune = yes
```
On both servers I run the following:

```
*/30 * * * * root TZ=UTC /usr/sbin/sanoid --cron
*/28 * * * * root TZ=UTC /usr/sbin/sanoid --monitor-snapshots && /usr/bin/curl -m 60 https://snitchendpointredacted
```

PR #272, opened 2018-09-13
Adds syncoid compress=zfs option to use zfs send -c

mattaw:
Based on b90cb849411a24e78ca13a32b8c2423b3838d44a from ckeasling/sanoid on 12th of December 2017.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1238,7 +1238,8 @@ syncoid - ZFS snapshot replication tool
 
 Options:
 
-  --compress=FORMAT     Compresses data during transfer. Currently accepted options are gzip, pigz-fast, pigz-slow, lzo (default) & none
+  --compress=FORMAT     Compresses data during transfer. Currently accepted options are zfs, gzip, pigz-fast, pigz-slow, lzo (default) & none
+  --compress=zfs        Uses zfs efficient built in compressed send option "zfs send -c" to directly send compressed blocks from source to target
   --identifier=EXTRA    Extra identifier which is included in the snapshot name. Can be used for replicating to multiple targets.
   --recursive|r         Also transfers child datasets
   --skip-parent         Skips syncing of the parent dataset. Does nothing without '--recursive' option.
--- a/syncoid
+++ b/syncoid
@@ -23,6 +23,11 @@ GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsn
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
 
+my $zfscompress = '';
+if (length $args{'compress'}) {
+  $zfscompress = $args{'compress'} eq 'zfs' ? '-c' : '';
+}
+
 # TODO Expand to accept multiple sources?
 if (scalar(@ARGV) != 2) {
 	print("Source or target not found!\n");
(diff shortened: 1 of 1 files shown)


mattaw:
I am wondering about the validity of this patch - it works but with zfs v7.5 on Ubuntu to make it work you need to do zfs send -Lce or I am reading the manual page wrong. That is to send large blocks and embedded blocks as well as compressed blocks. 
Thoughts? 
Better Options:
1) Add in -Lce? However we should probably check the zfs version?
2) Rewrite to allow a user new option, --zfssendflags where I could set -Lce myself, or any combination?

I prefer option 2 right now, rather than trying to test the zfs capabilities and switch things based on that. The user can set flags if they want to, or not.

syncoid
```diff
@@ -18,11 +18,16 @@ use Sys::Hostname;
...
 GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r",
-                   "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
+                   "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@", "sendoptions=s",
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s") or pod2usage(2);
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
 
+my $sendoptions = '';
+if (length $args{'sendoptions'}) {
+  $sendoptions = $args{'sendoptions'}
```
phreaker0: wrong indentation, sanoid uses tabs.

syncoid
```diff
@@ -379,7 +384,7 @@ sub syncdataset {
 		# and because this will ony resume the receive to the next
 		# snapshot, do a normal sync after that
 		if (defined($receivetoken)) {
-		    my $sendcmd = "$sourcesudocmd $zfscmd send -t $receivetoken";
+		    my $sendcmd = "$sourcesudocmd $zfscmd send $sendoptions -t $receivetoken";
```
phreaker0: this won't work mit for example -Lec as zfs -t only supports -e out of those flags.

syncoid
mattaw: Oh how embarrassing - my bad, I'll clean that up.

syncoid
mattaw: So just read through the zfs code, quite interesting. [Did you know there is an undocumented "raw" send available with -w? I certainly didn't ].

So zfs send -t -Lce would work just fine, because the resume token contains the flags from the original send and they will override any set with -t. -p would be rejected by the zfs utility if set with -t. 

However it clearly isn't correct (or sensible). I can remove the override from the zfs send -t path in syncoid to fix that properly. And allow -p in the original send, which will break right now.

mattaw:
Philosophically I'd rather this not be seen as just enabling the --compress=zfs concept. This has broader applicability, for my use cases zfs send -p as well as -Lce and I'll add another option to decorate zfs receive to enable -o property=value and -x options to enable different compression and properties on the receiving end.

Practically it would be too much work to carefully map out the allowable "send -Lce"/"receive -o property=value -o property=value -x property" options, create the test cases, detect OS, pool properties and possibly zfs versions. It could also introduce many bugs or corner cases if not done very carefully.

If given the chance I would rather allow this send and receive decoration (and document it) as something for experienced users only who will use the --debug command to check what happens.

I'll fix the zfs -t code path now I know zfs is keeping track of the send flags correctly in the token. I will also produce a patch to enable $recvoptions in syncoid for the first receive. Resuming a receive should preserve any -o property=value and -x property that was initially set on the first zfs receive of a failed send.

Would that be acceptable to the project? Much more powerful than just --compress=zfs but the user will have to handle breakage themselves. It preserves the power of syncoid for the oddballs like myself. If so I'll close this pull request and open one with a much more appropriate title.

syncoid
phreaker0: @mattaw yeah, but the raw send feature isn't in a stable release yet. It will be released together with the encryption feature in the upcoming 0.8 release to make secure remote backups possible.

You are right, despite what the manpag says, that combinations works.

PR #273, opened 2018-09-14
Add in poweruser --sendoptions=OPTIONS and --recvoptions=OPTIONS

mattaw:
Support poweruse use of things like "zfs send -Lce -p" and "zfs receive -o property=value -x property2" inside syncoid's great handing of resume, piping, etc.

I also converted spaces to tabs for internal consistency where appropriate.

Refer to #272 for discussion of an earlier version.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
 
 # this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved
-# from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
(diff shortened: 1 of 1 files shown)


nick4u:
I'd love to see this merged just to set --recvoptions="-x canmount" 
I do not need to see on backup server hundreds of lxc host container datasets  

I just looked at syncoid and I think there is mistake starting on line 32: 
```perl
my $recvoptions = '';
if (length $args{'recvoptions'}) {
	$sendoptions = $args{'recvoptions'}  # this is wrong? should be recvoptions = ... ?
}

jimsalterjrs:
superseded by @phreaker0's #295, thank you for your work @mattaw and @nick4u !

PR #280, opened 2018-10-16
add pre, post and prune snapshot scripts

### LINKED ISSUE #105: Feature Request: Pre/Post scripting

phreaker0:
implements pre/post script execution for snapshot creation. This replaces the PR #193 which was cherry picked, this pr adds:

- commit from @darkbasic for providing the snapshot name to the scripts
- timeout logic for the scripts so the can't hang sanoid forever
- cleanup of environment in any script error case

I still need to restructure the code a bit, make the timeout configurable and add documentation.

Fixes #105

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -455,12 +455,35 @@ sub take_snapshots {
 
 	if ( (scalar(@newsnaps)) > 0) {
 		foreach my $snap ( @newsnaps ) {
+			my $dataset = (split '@', $snap)[0];
+			my $presnapshotfailure = 0;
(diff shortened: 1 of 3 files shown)


darkbasic:
Good, I've created another PR against your repo to add pruning scripts: https://github.com/phreaker0/sanoid/pull/1

phreaker0:
@darkbasic yep, was going to, forgot to mention it in the description

phreaker0:
@jimsalterjrs can you have a look?
As this PR also changes the global defaults it should also get the version check from #157, right?

PR #286, opened 2018-11-09
removed sleeping between snapshot taking

phreaker0:
For me sanoid runtime is down by 1/3 with a small count of datasets and the difference will only get larger for a higher dataset count.
I don't see any problem if there are snapshots with the same creation time, in the typical use case the snapshots grouped by interval type won't have the same creation time and even if the will get the same creation time anyhow it shouldn't break anything.
But maybe I overlooked something, I will run this on several machines and check if anything comes up.


jimsalterjrs:
This one made me nervous at first because back in the dawn of time, I used creation time as part of the "is this the right snapshot" checks. We should long since have changed all that to look at UIDs, though, so... ::crosses fingers::

PR #287, opened 2018-11-14
split snapshot taking/pruning into seperate units for debian package …

phreaker0:
…to prevent pruning blocking snapshot taking if the later takes long time.

It happened to me on some occasions that the snapshot pruning blocked the taking, so it's better to do the pruning separately:

- i forgot the pruning configuration, ended up with ~100 000 datasets. After adding the correct pruning options, sanoid naturally ran a long time for pruning but a lot of snapshots were missed to be taken
- Received a huge amount of snapshots with a zfs received, which were pruned by sanoid delaying snapshot taking.


phreaker0:
this will also help with #289

PR #289, opened 2018-11-20
added option to defer pruning based on the available pool capacity

phreaker0:
Unused space is wasted space :-) With this PR a new option is introduced where one can defer any pruning of snapshots if less capacity is used as specified.
As this adds a new value, we probably want the default configuration version check from #157 before that.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -31,6 +31,7 @@ if (keys %args < 2) {
 my $pscmd = '/bin/ps';
 
 my $zfs = '/sbin/zfs';
+my $zpool = '/sbin/zpool';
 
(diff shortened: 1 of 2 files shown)


redmop:
I'm going to be honest. I'm a little turned on right now.

jimsalterjrs:
Ooh, love this. My first semi-sanoid-like backup application, which used rsync and hardlink-trees to do a file-based solution as close to snapshots as you could really get without a whole filesystem to support it, went by percentage of disk capacity available. I always kinda missed that a bit gently, but never did go through with re-implementing that approach as an additional layer in sanoid.

Whee!

PR #295, opened 2018-11-29
Added in --sendoptions=OPTIONS and --recvoptions=OPTIONS to inject OP…

clinta:
This is a cleaned up rebase of @mattaw's PR #273.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -17,12 +17,23 @@ use Sys::Hostname;
 # Blank defaults to use ssh client's default
 # TODO: Merge into a single "sshflags" option?
 my %args = ('sshkey' => '', 'sshport' => '', 'sshcipher' => '', 'sshoption' => [], 'target-bwlimit' => '', 'source-bwlimit' => '');
-GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r",
+GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r", "sendoptions=s", "recvoptions=s",
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
@phreaker0 moar rebase

clinta:
@jimsalterjrs @phreaker0 rebased again.

mattaw:
Thankyou so much, I got buried by the end of our semester and was planning to tackle it again near Christmas.

jimsalterjrs:
Still have conflicts that need to be resolved.

TerraTech:
I believe this needs a bit more fine tuning where some options need to be masked depending on the dataset being transferred.

Case in point is applying:  --recvoptions="-o canmount=noauto"
which will work fine for filesystem types, but will throw an error if a volume type is being processed.

I'm thinking that a possible solution is to extend the options to provide specificity, e.g.
- --recvoptions=
- --recvoptions-filesystem=
- --recvoptions-volume=

or if you don't want to balloon the number of options, then a additive building parser can be done:
- --recvoptions="<common to all dataset types>"
- --recvoptions=":filesystem:-o canmount=noauto"

where the applied options is scoped to the embedded ```:identifier:```  (I'm open to alternative delimiters)

The internal hash would be a HOL:
```perl
%recvoptions = (
  common => (...),
  filesystem => (...),
  volume => (...)
);
```

Then the correct options could be applied via introspection:
```perl
my $fstype = getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'type');
if ($fstype eq 'volume') {
  $receiveextraargs = join(' ', @recvoptions{common,volume});
}
elseif ($fstype eq 'filesystem') {
  $receiveextraargs = join(' ', @recvoptions{common,filesystem});
}
else {
  $receiveextraargs = join(' ', $recvoptions{common});
}
```

clinta:
@jimsalterjrs @phreaker0

Would it be possible to get this reviewed and merged before it falls out of sync and needs another rebase?

efinley1272:
+100

mailinglists35:
for my use case of "send -Lc" and "receive -vu" it does not break resumable replication

mailinglists35:
I see it correctly gets a resume token and if interrupted it correctly picks up the token and completes the send.

PR #306, opened 2018-12-16
support monitor time period suffixes

phreaker0:


jimsalterjrs:
@phreaker0 do you see any reason NOT to make the regex case-insensitive in your convert subroutine, so that 5w and 5W both refer to 5 weeks?

I'd like to get sanoid.defaults.conf to use explicit values (ie 5w instead of 5, 30h instead of 30, etc) once we've gotten some testing on this and are confident of it.

phreaker0:
@jimsalterjrs I actually had it that way until I came across the **M**onth and **m**inute issue :-) we could:
- have it case insensitive for all other suffixes
- use "mi" for minutes or "mo" for months
- leave it case sensitive

what do you prefer?

jimsalterjrs:
Oooh, excellent point. I'm in favor of (m)inutes and (M)onths, everything else can be case insensitive. Or just not having months be an option at all; I don't really believe there's much value in months as an option in this context. The only place months or weeks are really meaningful IMO is in taking monthlies or taking weeklies in the first place, and specifying how many to keep; but that's taken care of with "how many we keep" and therefore doesn't need a unit specified.

redmop:
I believe it should be all cast sensitive or none of it should be. For the end user.
Znapzend uses the following:
second|sec|s
minute|min
hour|h
day|d
week|w
month|mon|m
year|y

lckarssen:
Would it make sense to follow the conventions of the variables of the `date` tool? There, `%M` is for minute, `%m` is for month.

jimsalterjrs:
If simply dropping the month unit resolves the problem, I say we drop the month unit and resolve the problem that way. I just don't see any actual use for that unit, since "monthly = 3" captures three monthly snapshots without the need to either specify units or try to convert into days.

Issue #304, opened 2018-12-13
--no-resume required if snapshot removed after replication interrupted


jimsalterjrs:
If a replication is interrupted in a snapshot which is destroyed prior to the next attempt at replication, syncoid fails completely on that dataset unless --no-resume is specified, making the resume feature unsafe for unattended use.

Example:

````
Resuming interrupted zfs send/receive from data/images/fileserver to data/prod0/images/fileserver (~ UNKNOWN remaining):
cannot resume send: 'data/images/fileserver@autosnap_2018-09-14_13:00:01_hourly' used in the initial send no longer exists
0.00 B 0:00:00 [0.00 B/s] [<=>                                                                                     ]
cannot receive: failed to read from stream
CRITICAL ERROR: /usr/bin/ssh     -S /tmp/syncoid-root-root@cse-prod0-1544743400 root@cse-prod0 ' /sbin/zfs send -t 1-132fbc1035-110-789c636064000310a501c49c50360710a715e5e7a69766a63040017fa0ac627afc153905209b1d495d7e52566a7209030317541d867c5a5a716a0948262b5d9d0522cf86249f5459925a0ca4134a729660d35f920f7145f5e2b2bf19de961b2390e439c1f27989b9a90c0c29892589fa99b989e9a9c5fa699939a9c5a94565a9450e89a525f9c5798905f146068616ba0696ba8626f186c6560606560686f119f9a54539950c700000fabe2e01 | /bin/gzip -3 | /usr/bin/mbuffer -R 550K -q -s 128k -m 16M 2>/dev/null' | /usr/bin/mbuffer  -q -s 128k -m 16M 2>/dev/null | /bin/zcat  | /usr/bin/pv -s 0 |  /sbin/zfs receive -s -F 'data/prod0/images/fileserver' failed: 256 at /usr/local/bin/syncoid line 469.
````
@phreaker0, since you were the one who coded the resume functionality, do you mind taking a crack at fixing this? If ZFS can't resume due to lack of the missing snapshot, syncoid should fall back to --no-resume behavior from the most recent remaining common snapshot.

This is a pretty easy bug to hit for anybody with Sanoid hourlies and daily offsite replication; get interrupted in the middle of an hourly, wait a day, try again... sanoid's already pruned the hourly, and boom, replication is broken until manual intervention.

phreaker0:
[...]

PR #309, opened 2018-12-19
WIP: reset resume state

phreaker0:
Relates to #304. This is the rebased attempt from some time ago, it detects a partial resume state and resets it so it will work but I'm not happy because it hides the command output. I need to find a way to:

- run the sync command like system()
- but with support for capturing the stderr output
- while the stdout and stderr are still printed to the user 

So more error types can be captured and handled properly (like invalid resume state, ...)


jimsalterjrs:
Hey @phreaker0 I just had a thought about this. While we do still need to manage broken resume states in syncoid... how hard is it to *check* for a resume state? Should we look at having Sanoid refuse to prune a snapshot with an associated resume state? Is that practical? If so, it would eliminate 95%+ of the actual real-world problems associated with this issue.

Honestly I kinda wonder if we shouldn't be reporting this upstream; zfs itself should probably refuse to destroy a snapshot with an associated resume state without use of a -F flag, much like it won't destroy a snapshot associated with a clone without one.

phreaker0:
@jimsalterjrs 
The resume state is only saved on the destination dataset, so a sanoid run on the src dataset/snapshot unfortunately can't detect snapshots which are part of an partial replication.

jimsalterjrs:
Is this perhaps worth defining replication partners in sanoid.conf, 
allowing a check of the replication partners for existing resume states? Or 
is checking for resume states too heavy an operation?

phreaker0:
This seems to be more complicated than a proper solution which should work without configuration and hopefully in all cases :-)

jimsalterjrs:
@phreaker0 have you done any testing on this commit yet?

phreaker0:
@jimsalterjrs yes it works. I can get the error output from the zfs command while it's also directly forwarded to the terminal. But the only caveat is the need of an additional perl package (Captrue::Tiny). I still need to rewrite it to be nicer but like this I can catch other errors too easily and can handle them automatically instead of aborting if possible.

jimsalterjrs:
Freaking sweet; really looking forward to being able to turn off --no-resume on my automated cron jobs!

phreaker0:
You really wanted that feature :-D I will push some documentation updates regarding this later.

PR #312, opened 2018-12-20
preserve taking snapshots order always from yearly to frequently && pre/post snapshot script calls should be also dumped in --readonly mode

matveevandrey:


matveevandrey:
As I wrote in my commit this double loop is ugly but the simplest way to achieve required with minimal changes. It **can not**  be done "directly iterate over the known types" because the key should also be found to get corresponding values

phreaker0:
The key always exists for all the known types because they are defined in the sanoid.defaults.conf file which shouldn't be changed by the user and is part of the installation.

matveevandrey:
done

phreaker0:
thank you! I still need to try it out.

PR #319, opened 2019-01-07
removed debug output from excludes loop

### LINKED ISSUE #314: Syncoid outputs not filtered list in recursive mode
when running syncoid with --recursive and --exclude= argument it always prints ALL child datasets, without applying exclude filter, but replicates correctly.

phreaker0:
Fixes #314


phreaker0:
this is actually just a simple removal of debug output

PR #320, opened 2019-01-07
zfs recursion

### LINKED ISSUE #104: Feature Request: Make snapshots atomic relative to one another
I use multiple virtual disks per VM/CT. They are in the same tree (rpool/data/vms/snap-15/vm-100-disk-[1|2] for example). Sanoid takes snapshots one at a time, so disk-1 is not atomic with disk-2

The new syncoid "--no-sync-snap" option takes care of part of it.

### LINKED ISSUE #173: Feature Request: Transaction / consistancy groups.
I have a single application that has 4 volumes I replicate to a 2nd location. I would like the snapshots to be closer together before the replication begins. Today a have a script that runs and syncs one after the other, which means the last one could be some time behind the first one as replication takes some time to complete.

The feature would basically iterate the list of volumes passed to syncoid and create the volume snapshots for each first, creating the initial snaps seconds from each other, and then replicate that data to the replication location so the application would be more consistent on the copy.

The list could be a list passed via command line (just keep passing source,targets) or maybe in a text file with a list of source/targets line by line.

May want to consider doing a transaction group so if the 4th replication fails, the transaction rolls back all the syncs rather than just fails the 1st one.

phreaker0:
this is hopefully a fixed version of the PR #181:

- i kept the "recursive" key in the toggle list to keep backward compatibility, but documented the reason and special value
- fixed the uninitialised variable error
- added warning if the user tries to force different snapshot settings on datasets below zfs recursions
- still allow ordinary top level recursion to apply other settings (monitoring, pruning, ...)

Fixes #173
Fixes #104

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -502,7 +502,13 @@ sub take_snapshots {
 					# update to most current possible datestamp
 					%datestamp = get_date();
 					# print "we should have had a $type snapshot of $path $maxage seconds ago; most recent is $newestage seconds old.\n";
-					push(@newsnaps, "$path\@autosnap_$datestamp{'sortable'}${dateSuffix}_$type");
+
+					# use zfs (atomic) recursion if specified in config
+					if ($config{$section}{'zfs_recursion'}) {
+						push(@newsnaps, "$path\@autosnap_$datestamp{'sortable'}${dateSuffix}_$type\@");
+					} else {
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
@phreaker0 - is this documented in README.md yet?

PR #342, opened 2019-01-30
implemented creation of zfs bookmarks

### LINKED ISSUE #336: feature request: syncoid should also create bookmark of the source snapshot that it finished sending
I am using external tool to create snapshots at source, different than sanoid (zfs-auto-snapshot, because it has the ability to skip creation of snapshots if their size would be zero).

I am also doing unfrequent sends to remote using syncoid, and sometimes source snapshots get rotated so there is no more common snapshot left between source and target.

When I originally filed https://github.com/jimsalterjrs/sanoid/issues/88, I tought I could solve the issue of deleted snapshots from source.
It turns out that #88 does not solve my goal, though: *somebody* must create the bookmark, too - in my case, I should take my chances to modify that shell script to also take a bookmark along with the snapshot.

But before that I would like to ask if it is possible to make syncoid take a bookmark of the source snapshot as well when invoked with `--no-sync-snap` **and** there is a special option `--create-bookmark` present (and create it only after a successfull send, or else the bookmark would be of no practical use), and at the same time never delete it so it can increase the chances of finding the corresponding snapshot at target when scanning for bookmark fallback.

Having an huge/unlimited number of bookmarks will not hurt anyway, and user should decide if it is worth to delete them.

**The advantage of having syncoid create the bookmarks it that it would allow syncoid to function independent of any other third party snapshot management solution, at practically no cost to the zfs datasets and not breaking any other scripts.**

I am practically illiterate with perl, so I attempted to monkey copy paste some snapshot creation code but it miserably fails.

What I did was to clone the area where `system($synccmd)` occurs with something like:
````
my $hostid = hostname();
my %date = getdate();
my $bookname = "synbook\_$identifier$hostid\_$date{'stamp'}";
my $takebookmark = "$sourcesudocmd $zfscmd bookmark $sourcefsescaped\@$oldestsnapescaped $sourcefs\#$bookname";

system($takebookmark) == 0 or do {
       warn "CRITICAL ERROR: bookmark creation failed: $?";
       return 0;
}
````

but I ran into a bunch of perl errors and I've given up.

Is there any chance this can be something trivial for you to do?

Many thanks!

phreaker0:
Fixes #336

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,10 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 	This argument tells syncoid to restrict itself to existing snapshots, instead of creating a semi-ephemeral syncoid snapshot at execution time. Especially useful in multi-target (A->B, A->C) replication schemes, where you might otherwise accumulate a large number of foreign syncoid snapshots.
 
++ --create-bookmark
+
+	This argument tells syncoid to create a zfs bookmark for the newest snapshot after it got replicated successfully. The bookmark name will be equal to the snapshot name. Only works in combination with the --no-sync-snap option. This can be very useful for irregular replication where the last matching snapshot on the source was already deleted but the bookmark remains so a replication is still possible. 
+
 + --no-clone-rollback
 
 	Do not rollback clones on target
--- a/syncoid
+++ b/syncoid
@@ -682,7 +682,7 @@ sub syncdataset {
 		if (defined $args{'create-bookmark'}) {
 			my $bookmarkcmd;
 			if ($sourcehost ne '') {
-				$bookmarkcmd = "$sshcmd $sourcehost " . escapeshellparam("$zfscmd bookmark $sourcefsescaped\@$newsyncsnapescaped $sourcefsescaped\#$newsyncsnapescaped");
+				$bookmarkcmd = "$sshcmd $sourcehost " . escapeshellparam("$sourcesudocmd $zfscmd bookmark $sourcefsescaped\@$newsyncsnapescaped $sourcefsescaped\#$newsyncsnapescaped");
 			} else {
 				$bookmarkcmd = "$sourcesudocmd $zfscmd bookmark $sourcefsescaped\@$newsyncsnapescaped $sourcefsescaped\#$newsyncsnapescaped";
 			}
--- a/syncoid
+++ b/syncoid
@@ -23,6 +23,7 @@ GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsn
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
                    "no-clone-handling", "no-privilege-elevation", "force-delete", "no-clone-rollback", "no-rollback",
+                   "create-bookmark",
                    "mbuffer-size=s" => \$mbuffer_size) or pod2usage(2);
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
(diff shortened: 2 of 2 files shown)


mailinglists35:
I have one more case :)

Let's have this:

no target
`[suport@zfspv ~]$ zfs get -r guid bkppv/encrypted/bkphq/delete
cannot open 'bkppv/encrypted/bkphq/delete': dataset does not exist`

source with one manual snapshot
`[suport@zfshq ~]$ zfs get guid -r bkphq/delete
NAME                    PROPERTY  VALUE  SOURCE
bkphq/delete            guid      18092307491229082710  -
bkphq/delete@firstsync  guid      2836662525218684215  -
`
invoke first sync with `--debug --create-snapshot --no-sync-snap` https://pastebin.com/aYESSPiU
which will create target and send from the only source snapshot, then will create bookmark

invoke a second time, and this time (output included in pastebin above too) after "INFO: no snapshots on source newer than firstsync on target. Nothing to do, not syncing.", it attempts to create the bookmark, but it fails because it has the same name.

I was wondering if wether
should this give up creating bookmark after it finds identical latest snapshot on source and target, or should it delete the matching bookmark and create it again to match the matching snapshot? 

but then I thought that later in time we can be in the following situation:
original matching snapshot can get deleted at source, but later a new one created with the same name

So neither recreating nor abandoning creation would be optimal, but creating the bookmark with an additional suffix in the name (for example -$(seconds since midnight padded with zeroes to be a fixed length)-$(random 3 digits) - i could be wrong but that's the first that i imagined EDIT: or first field of cksum coreutils command with the snapshot's guid at input - `zfs get guid -o value -H bkphq/delete@firstsync | cksum | cut -d' ' -f1` ). Since when searching matching pairs we are comparing guids, this will not affect finding the matching bookmark-snapshot pair, but will allow creating new bookmark even if the snapshot name is the same but taken at different time with a different guid than the original one. I would not delete the bookmark since it might be useful when replicating to a second or third target.

What do you think?

mailinglists35:
if that complicates things too much or takes too much of your time, i'd be happy even with random bookmark name instead, ie:

`$bookmarkcmd = "$sourcesudocmd $zfscmd bookmark $sourcefsescaped\@$newsyncsnapescaped $sourcefsescaped\#$<guid of $newsyncsnap>";`

mailinglists35:
oh well, I think both take about the same time and effort... since in both cases you'd do an additional `$bookname=$( something )` then use `$bookname` instead of `$newsyncsnapescaped` in `$bookmarkcmd`

mailinglists35:
even with pseudo unique bookmark names we still must perform an additional check when creating, so we do not attempt to create the same one again. i'm trying to look around $bookmarkcmd to write a PR with this verify

mailinglists35:
sorry, i've done quite a mess both in perl and github but it works for me now.
I hope your perl knowledge can transform my miserable coding into a better one :)

https://github.com/phreaker0/sanoid/pull/2

mailinglists35:
I got this running and it works pretty well - thank you so much!

I've looked at some cases where source and target were linked by just one single common snapshot while each having lots of other snapshots; the sync process sent lots of intermediary snapshots, but the bookmark creation was done only for the newest snapshot.

Is it possible to create bookmarks for each intermediary snapshot that is found to be between oldest snapshot and newest snapshot? This would greatly increase future chances of falling back to a bookmark by having more bookmarks available at source

Thank you again and have a great day!

phreaker0:
@mailinglists35 this can be implemented in a follow up PR

ju1m:
Makin `syncoid` create intermediary snapshots **would be a great addition**.
I wrongly assumed that `syncoid` did that and it has bitten me because I'm using `hourly` snapshots (using `sanoid`) which makes `syncoid` always sending `hourly` snapshots and thus **only creating `hourly` bookmarks**, but because I'm pruning `hourly` snapshots older than a day, the first time I became without a common `hourly`-or-less-frequent snapshot I was also without a common bookmark.

Xasin:
Apologies for replying to this closed PR, but I wanted to ask why, specifically, Bookmarks are never deleted?
I do understand that they don't have much of a footprint at all, but I also think that only the latest bookmark, created by Syncoid itself, is useful, and older bookmarks couldn't be used for syncing anyways.

So, why keep them around? They would eventually clutter up, potentially getting in the way.
Perhaps the bookmarks could be tagged per syncoid "target", like Sync snaps would be, and each syncoid target deletes the old bookmark after successfully creating the new bookmark.

phreaker0:
think of the following scenario:

production pool "src" which is backuped irregularly to an offsite "dst" pool.

* snapshot1-99 on src pool (each snapshot with a created bookmark)
* replication to dst pool (snapshot1-99 now exist on dst pool)
* snapshot1-99 was pruned from src pool
* snapshot300-400 exist on src pool
* replication to dst pool
  * no common snapshot anymore
  * bookmark of snapshot99 matches dst snapshot99
  * replication can take place from bookmark of snapshot99

you suggested to keep only the latest bookmark which syncoid creates, in this case this would be bookmark for snapshot400 and in this case replication wouldn't be possible anymore.
The cleanup of bookmarks is highly dependent on your use case and should be done outside of syncoid IMO.

PR #351, opened 2019-02-17
Sanoid ebuild for Gentoo

thehaven:
A quick and simple ebuild for Gentoo for anyone wishing to use Sanoid on that OS.


ste7677:
...the files should be moved from gentoo/sys-fs/* to gentoo/sys-fs/sanoid/*

the structure on thehaven's gentoo overlay is correct, see [https://github.com/thehaven/haven-overlay/tree/master/sys-fs](url)

thehaven:
Thanks, you are indeed correct - sorry I missed that first time round. New pull request to fix is at:
* https://github.com/jimsalterjrs/sanoid/pull/363

PR #352, opened 2019-02-19
use ordinary replication if clone recreation fails

### LINKED ISSUE #323: recursive clones can fail on the first replication attempts
Depending on the order of recursive clones they will fail on the first attempts because the origin datasets weren't replicated before. Syncoid should order the datasets for replication so the origin of all cones are replicated before recreating them.

phreaker0:
Fixes #323

It just uses the no clone replication as fallback if clone creation fails. In the future I would like to do this with proper clone creation handling and detection code so every clone is recreated if possible.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -416,6 +416,12 @@ sub syncdataset {
 			return 0;
 		}
 		system($synccmd) == 0 or do {
+			if (defined $origin) {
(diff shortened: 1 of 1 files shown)


phreaker0:
i tested the patch and use it in production.

PR #367, opened 2019-03-31
Revert #365

phreaker0:
This PR reverts #365 which removed the sanoid-prune.service from the install instruction.
@danielewood It is not referenced by a timer but it's referenced by the main sanoid.service and is run afterwards. For an explanation why this is a better approach see this PR #287.

I kept the --verbose flag as it's useful for journald logging.


danielewood:
Sounds good to me. I'll test it on Monday to see if it triggers.

PR #368, opened 2019-04-02
Added line to enable sanoid-prune.service

danielewood:
@phreaker0,
I found that on my systems(CentOS7), sanoid-pune.service did nothing until I enabled it. Once I did that, it was called by sanoid.service.

References #367


phreaker0:
thx! i missed that (the debian package automatically enables the service).

PR #362, opened 2019-03-26
Changed (and simplified) DST handling

shodanshok:
See https://github.com/jimsalterjrs/sanoid/issues/155 for background information.

This patch streamlines `sanoid` DST handling, reverting the "snapshot suffix" behavior introduced with https://github.com/jimsalterjrs/sanoid/pull/175. 

The only downside of this approach is the missing of a single hourly snapshot (with duplicate 02:00:00 time). To avoid that, one should use UTC time (as clearly documented).

PR #369, opened 2019-04-03
Dont force TZ=UTC for sys-fs/sanoid cron on gentoo

thehaven:
Removing TZ=UTC from cron for the gentoo build as thats causing some timezone confusion for some folks.


phreaker0:
@thehaven 
Removing this will probably cause more confusion:
- currently for the DST day hourly snapshots will get a "_y" suffix
- if #362 get's merged one hourly snapshot will be lost on the DST day

I think the UTC behavior is therefore better as a default and I also would dislike if the packaging defaults would differ (deb, rpm, source, ...).

phreaker0:
@jimsalterjrs do you have a opinion on this?

phreaker0:
#362 got merged and because of the stated issues I rather don't change the default here.

PR #374, opened 2019-04-09
Do not require user to be specified for syncoid

aerusso:
As it stands, syncoid does not allow ssh targets that do not specify a user. This patch adds that feature, requiring several changes to its behavior:

1. We bring the behavior of syncoid into line with the documentation, which suggests that `:`, rather than `@`, is required when specifying a remote dataset.
2. The behavior of a TARGET or SOURCE that begins with `:` is altered to allow for local dataset names with colons in them.
3. The name of the control socket is changed slightly.

I've been using a version of this patch for around a year. I'm currently using it (rebased onto v2.0.1). The patch also includes a note in the help describing the change.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1420,20 +1420,53 @@ sub targetexists {
 sub getssh {
 	my $fs = shift;
 
-	my $rhost;
+	my $rhost = "";
 	my $isroot;
 	my $socket;
+	my $remoteuser = "";
 
 	# if we got passed something with an @ in it, we assume it's an ssh connection, eg root@myotherbox
 	if ($fs =~ /\@/) {
 		$rhost = $fs;
-		$fs =~ s/^\S*\@\S*://;
+		$fs =~ s/^[^\@:]*\@[^\@:]*://;
 		$rhost =~ s/:\Q$fs\E$//;
-		my $remoteuser = $rhost;
-		 $remoteuser =~ s/\@.*$//;
+		$remoteuser = $rhost;
+		$remoteuser =~ s/\@.*$//;
+		# do not require a username to be specified
+		$rhost =~ s/^@//;
+	} elsif ($fs =~ m{^[^/]*:}) {
+		# if we got passed something with an : in it, BEFORE any forward slash
+		# (i.e., not in a dataset name) it MAY be an ssh connection
+		# but we need to check if there is a pool with that name
+		my $pool = $fs;
+		$pool =~ s%/.*$%%;
+		my ($pools, $error, $exit) = capture {
+			system("$zfscmd list -d0 -H -oname");
+		};
+		$rhost = $fs;
+		if ($exit != 0) {
+			warn "Unable to enumerate pools (is zfs available?)";
+		} else {
+			foreach (split(/\n/,$pools)) {
+				if ($_ eq $pool) {
+					# there's a pool with this name.
+					$rhost = "";
+					last;
+				}
+			}
+		}
+		if ($rhost ne "") {
+			# there's no pool that might conflict with this
+			$rhost =~ s/:.*$//;
+			$fs =~ s/\Q$rhost\E://;
+		}
+	}
+
+	if ($rhost ne "") {
 		if ($remoteuser eq 'root' || $args{'no-privilege-elevation'}) { $isroot = 1; } else { $isroot = 0; }
 		# now we need to establish a persistent master SSH connection
-		$socket = "/tmp/syncoid-$remoteuser-$rhost-" . time();
+		$socket = "/tmp/syncoid-$rhost-" . time();
+
 		open FH, "$sshcmd -M -S $socket -o ControlPersist=1m $args{'sshport'} $rhost exit |";
 		close FH;
 
@@ -1882,9 +1915,9 @@ syncoid - ZFS snapshot replication tool
 =head1 SYNOPSIS
 
  syncoid [options]... SOURCE TARGET
- or   syncoid [options]... SOURCE USER@HOST:TARGET
- or   syncoid [options]... USER@HOST:SOURCE TARGET
- or   syncoid [options]... USER@HOST:SOURCE USER@HOST:TARGET
+ or   syncoid [options]... SOURCE [[USER]@]HOST:TARGET
+ or   syncoid [options]... [[USER]@]HOST:SOURCE TARGET
+ or   syncoid [options]... [[USER]@]HOST:SOURCE [[USER]@]HOST:TARGET
 
  SOURCE                Source ZFS dataset. Can be either local or remote
  TARGET                Target ZFS dataset. Can be either local or remote


redmop:
I'm just curious. What is the use case? What problem does this solve that not putting the username in there required?

aerusso:
In my case, I would like to avoid duplicating information about host configuration (DNRY). I specify host configuration information in one place (~/.ssh/config) and manage it there.

It's also nice because it brings syncoid's behavior in line with ssh and scp (which also infer username information, in precisely this way).

I'll also point out that [dataset names must begin with an alphanumeric character](https://docs.oracle.com/cd/E26505_01/html/E37384/gbcpt.html), so this patch cannot break any valid, existing syncoid call. It's completely backwards compatible.

In my mind, the question is: why do we *require* a username?

phreaker0:
@aerusso What about handling an empty ssh host as local replication, this way : colon can still be used locally but with an extra : in front:
`syncoid :pool/src:data pool/dst`

aerusso:
That's actually the current behavior of this patch. Existing configurations
with local datasets with colons in a name will be broken, though. The
resolution would be to put a colon at the front, as you suggest.

phreaker0:
@aerusso nice, i overlooked that.

jimsalterjrs:
I don't like "datasets with a colon in the name get broken", even with workarounds. I don't like the front-loaded empty colon, either.

If you're really hell-bent on not having to specify username, I'd maybe prefer allowing `syncoid pool/src @:remotebox:pool/target`, which would have the virtue of not breaking anybody's stuff or requiring any additional symbols.

aerusso:
@jimsalterjrs I looked back at this code, and realized that there is actually no issue with the `:` character---pool names must not contain them. So, this newest revision checks if there is a `:` BEFORE any `/` characters, and then proceeds.

So, there's actually no changes needed to the documentation, and no breakage to any existing configurations!

This has not been as tested, though---I just wrote it up---so a review is definitely something I'd support.

aerusso:
@phreaker0 Did you test with the most recent version? It should not break any existing configuration.

phreaker0:
@aerusso yes, my reproducer:

```
POOL_IMAGE="/tmp/syncoid-test-8.zpool"
MOUNT_TARGET="/tmp/syncoid-test-8.mount"
POOL_SIZE="1000M"
POOL_NAME="syncoid-test-8-:"

truncate -s "${POOL_SIZE}" "${POOL_IMAGE}"

zpool create -m none -f "${POOL_NAME}" "${POOL_IMAGE}"

function cleanUp {
  zpool export "${POOL_NAME}"
}

# export pool in any case
trap cleanUp EXIT

zfs create "${POOL_NAME}"/src
zfs create "${POOL_NAME}"/src/" -_.:"cool" :._- "
../../../syncoid --recursive --debug --compress=none "${POOL_NAME}"/src "${POOL_NAME}"/dst
sleep 1
../../../syncoid --recursive --debug --compress=none "${POOL_NAME}"/src/" -_.:"cool" :._- " "${POOL_NAME}"/dst/" -_.:"cool2" :._- "
sleep 1
../../../syncoid --recursive --debug --compress=none "${POOL_NAME}"/src/" -_.:"cool" :._- " root@127.0.0.1:"${POOL_NAME}"/dst/" -_.:"cool3" :._- "
sleep 1
../../../syncoid --recursive --debug --compress=none "${POOL_NAME}"/src/" -_.:"cool" :._- " 127.0.0.1:"${POOL_NAME}"/dst/" -_.:"cool4" :._- "

zfs list -r "${POOL_NAME}"
```

aerusso:
That's an illegal pool name---you're [not supposed to have colons in it](https://docs.oracle.com/cd/E26505_01/html/E37384/gbcpt.html).

phreaker0:
Seem like this is solaris specific. The example code from above worked on my machine and from the manpage of zpool (linux zfs 0.8.2) it also documented to be allowed:

```
...The pool name must begin with a letter, and can only contain alphanumeric characters
as well as underscore ("_"), dash ("."), colon (":"), space ("-"), and period ("."). 
The pool names mirror, rai....
```

aerusso:
Yeah, I noticed that in the edit. This actually breaks syncoid, not just this patch.

phreaker0:
`syncoid --recursive --debug --compress=none "${POOL_NAME}"/src "${POOL_NAME}"/dst`
from the example above worked for me on linux without this PR.

aerusso:
Sorry, you are correct. It's too early here.

phreaker0:
@aerusso :-) so what's your opinion about the proposed syntax:

`syncoid localSrcPool @remoteServer:remoteDstPool`

aerusso:
I'm planning on using that syntax, but I've got one more trick I want to put in to avoid unnecessary requirement of the `@`.

aerusso:
@phreaker0 This resolves the ambiguity in the pool name by checking for any possible local pool name conflicts.

You can also always use the `@` syntax.

Let me know how you feel about this version.

phreaker0:
@aerusso this looks nice! One case which currently doesn't work (and didn't before) is:

POOL_NAME="syncoid-test-8-:"
syncoid --recursive --debug --compress=none "${POOL_NAME}"/src/" -_.:"cool" :._- " root@127.0.0.1:"${POOL_NAME}"/dst/" -_.:"cool3" :._- 

While you are at it, can you also use the same handling you implemented for not specified usernames where the same command works.

aerusso:
>  One case which currently doesn't work (and didn't before) is [snip]

It should work now, in this revision. Thanks for checking these, by the way!

> While you are at it, can you also use the same handling you implemented for not specified usernames where the same command works.

I'm a little confused about what you mean here. Could you give me an example what you'd like changed?

phreaker0:
@aerusso thx, I will test this later. It was about the same thing, sorry for the wording :-)

syncoid
```diff
@@ -1420,20 +1420,49 @@ sub targetexists {
...
+		$rhost =~ s/^@//;
+	} elsif ($fs =~ m{^[^/]*:}) {
+		# if we got passed something with an : in it, BEFORE any forward slash
+		# (i.e., not in a dataset name) it MAY be an ssh connection
+		# but we need to check if there is a pool with that name
+		my $pool = $fs;
+		$pool =~ s%/.*$%%;
+		my ($pools, $error, $exit) = capture {
+			system("$zpoolcmd list -H -oname");
+		};
```
phreaker0: The exit code should be checked here

phreaker0:
@aerusso I tested the change and it works, thx! One last thing, can you add a check for the exit code of the zpool list command and error out if there was a problem.

aerusso:
@phreaker0 I've pushed a version that checks the exit code, and warns on an error. I've also rewritten it to use `zfs` instead of `zpool`, so that users not running as root don't have to also add `zpool` to their path.

phreaker0:
@aerusso thx! all my test cases worked.

aerusso:
I just re-based this.  @jimsalterjrs Is there anything else you'd like me to do for this?

PR #375, opened 2019-04-20
Declare that sanoid service 'Wants' sanoid-prune

johnramsden:
The `sanoid-prune.service` never runs after `sanoid.service`.

Declare a `Wants` dependency, and `Before` condition on `sanoid-prune.service` in the sanoid unit, and prune is run after 'sanoid.service'.

The `WantedBy` condition currently in the `sanoid-prune.service` unit will not apply to oneshot units in the way required because conditions in the install section are only applied upon enable/disable.

From [systemd.unit](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#%5BInstall%5D%20Section%20Options) `[Install]` Section Options
> Unit files may include an "[Install]" section, which carries installation information for the unit. This section is not interpreted by systemd(1) during runtime; it is used by the enable and disable commands of the systemctl(1) tool during installation of a unit.

---

Also add --verbose flag for logging to journald, as mentioned in readme.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -78,6 +78,8 @@ cat << "EOF" | sudo tee /etc/systemd/system/sanoid.service
 Description=Snapshot ZFS Pool
 Requires=zfs.target
 After=zfs.target
+Wants=sanoid-prune.service
+Before=sanoid-prune.service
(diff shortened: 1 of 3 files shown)


phreaker0:
Didn't know that, tested it and it works.

PR #382, opened 2019-05-15
fixed ordering of snapshots with the same creation timestamp

### LINKED ISSUE #381: syncoid not always getting oldest snapshot
Discovered a bug today in which a full replication isn't actually beginning with the very oldest snapshot present:

    root@SAP0:~# zfs list -rt snap data/images | grep images\@
    data/images@autosnap_2019-05-08_13:13:56_monthly               0B      -    96K  -
    data/images@autosnap_2019-05-08_13:13:56_daily                 0B      -    96K  -
    data/images@autosnap_2019-05-08_13:13:56_hourly                0B      -    96K  -
    data/images@autosnap_2019-05-08_14:00:02_hourly                0B      -    96K  -
    data/images@autosnap_2019-05-08_15:00:01_hourly                0B      -    96K  -
    data/images@autosnap_2019-05-08_16:00:02_hourly                0B      -    96K  -
    data/images@syncoid_usp-hs0_2019-05-08:16:51:28                0B      -    96K  -

Replicating in full to hotspare misses the monthly snapshot:

    root@hs0:~# syncoid root@sap0:data/images data/sap0/data/images
    WARN: ZFS resume feature not available on target machine - sync will continue without resume support.
    INFO: Sending oldest full snapshot data/images@autosnap_2019-05-08_13:13:56_hourly (~ 40 KB) to new target filesystem:
      48.3KiB 0:00:00 [1.64MiB/s] [=================================] 118%
    INFO: Updating new target filesystem with incremental data/images@autosnap_2019-05-08_13:13:56_hourly ... syncoid_usp-hs0_2019-05-08:16:54:09 (~ 33 KB):
      14KiB 0:00:00 [ 321KiB/s] [=============>                    ] 42%

Note that it incorrectly identified the hourly, not the monthly, as "oldest snapshot" on the source. This means that we miss the monthly on the target:

    root@hs0:~# zfs list -rt snap data/sap0
    NAME                                                            USED  AVAIL  REFER  MOUNTPOINT
    data/sap0/data/images@autosnap_2019-05-08_13:13:56_hourly   144K      -   208K  -
    data/sap0/data/images@autosnap_2019-05-08_14:00:02_hourly    16K      -   208K  -
    data/sap0/data/images@autosnap_2019-05-08_15:00:01_hourly    16K      -   208K  -
    data/sap0/data/images@autosnap_2019-05-08_16:00:02_hourly    16K      -   208K  -
    data/sap0/data/images@syncoid_usp-hs0_2019-05-08:16:51:28    16K      -   208K  -
    data/sap0/data/images@syncoid_usp-hs0_2019-05-08:16:54:09      0      -   208K  -

Hey @phreaker0 , remember how I used to have `sanoid` sleep for one second between snapshots? *This* is why I was doing it. I remembered I had problems not being able to identify the oldest snapshot but I couldn't dredge up the details; now the details have re-bitten me in the arse and I remember why I was doing that. =)

When several snapshots of the same dataset are all the same age, it seems to be basically random which one gets pulled up as "oldest". I'm not sure if we're currently sorting age by the actual zfs creation property, or just using the text datestamp in the snapshot name itself. But neither one has a granularity smaller than one second, so it's easy to get order screwed up if you have multiples with the exact same creation time (to a granularity of one second). Suggestions that are less crude than my old one of "just sleep a second between taking snapshots, so this crap doesn't happen"? =)

phreaker0:
Fixes #381

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1393,6 +1393,8 @@ sub getsnaps() {
 	# this is a little obnoxious. get guid,creation returns guid,creation on two separate lines
 	# as though each were an entirely separate get command.
 
+	my %creationtimes=();
+
 	foreach my $line (@rawsnaps) {
 		# only import snap guids from the specified filesystem
 		if ($line =~ /\Q$fs\E\@.*guid/) {
@@ -1413,7 +1415,24 @@ sub getsnaps() {
 			$creation =~ s/^.*\tcreation\t*(\d*).*/$1/;
 			my $snap = $line;
 			$snap =~ s/^.*\@(.*)\tcreation.*$/$1/;
-			$snaps{$type}{$snap}{'creation'}=$creation;
+
+			# the accuracy of the creation timestamp is only for a second, but
+			# snapshots in the same second are highly likely. The list command
+			# has an ordered output so we append another three digit running number
+			# to the creation timestamp and make sure those are ordered correctly
+			# for snapshot with the same creation timestamp
+			my $counter = 0;
+			my $creationsuffix;
+			while ($counter < 999) {
+				$creationsuffix = sprintf("%s%03d", $creation, $counter);
+				if (!defined $creationtimes{$creationsuffix}) {
+					$creationtimes{$creationsuffix} = 1;
+					last;
+				}
+				$counter += 1;
+			}
+
+			$snaps{$type}{$snap}{'creation'}=$creationsuffix;
 		}
 	}
 


jimsalterjrs:
Thank you for this. But, correct me if I'm wrong, this only works if you assume that the same instance of sanoid creates all timestamps taken in the same second, right? This is *likely* but not guaranteed. Wouldn't it make more sense to use a more granular timestamp instead of a counter, ie go down to microsecond timing?

Also - are we certain that none of the regexes parsing out sanoid snapshots are looking for the format without the newly appended utime or counter, and would fail when we do append them?

phreaker0:
No, it works regardless. Every timestamp is suffixed internally with "000" when parsing the zfs list output until there are duplicates in which case the counter is used to preserve ordering. (The limit are 1000 snapshots per second) The proper solution would be to retrieve the creation time in millisecond precision but I don't  think zfs offers that as far as I found, therefore the workaround with the counter.

I'm not quiet sure what you mean with the regex parsing, the suffixed timestamp is only used internally and never written anywhere/read in.

jimsalterjrs:
Oh, okay - I misunderstood the impact. Thanks for explaining!

PR #383, opened 2019-05-16
Gracefully handle error when source dataset disappeared

### LINKED ISSUE #380: syncoid: Ignore ZFS filesystems that went away while running
If a dataset that is to be backed up goes away while `syncoid` is running, then the run ends with an error, even if the dataset was ignored.

For me, how this manifests is that I am using `syncoid` with ZFS on Linux, running docker on volume `zroot/var/docker`.  For the unaware, the `docker` ZFS driver creates a **LOT** of ZFS filesystems if you are working with a significant number of containers. I don't care to back these up (there are hundreds of filesystems under here on my systems) so I have `syncoid:sync` set to `false` on `zroot/var/docker`.

But if docker removes one of the datasets underneath `zroot/var/docker` during the run, it will produce an error like:

```
cannot open 'zroot/var/docker/ea93f6fc82682a94a21f1b91a6e66eb5a04378a9563e5f3f7229d0117c9327ee': dataset does not exist
cannot open 'zroot/var/docker/ea93f6fc82682a94a21f1b91a6e66eb5a04378a9563e5f3f7229d0117c9327ee-init': dataset does not exist
```
It would be nice if `syncoid` was smart enough to ignore these since `zroot/var/docker` already has `syncoid:sync=false`.  I'd be okay if this was non-default and triggered by a CLI flag also like `--ignore-missing-datasets` or something.

mschout:
If the source dataset dissappeared before we were able to get the
sanoid:sync property, do not set an error exit code.  Handle this
gracefully as this should not be considered an error.

Fixes #380

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1174,7 +1174,6 @@ sub dataset_exists {
 	close $fh;
 
 	if ($result =~ /dataset does not exist/) {
-		warn "$fsescaped does not exist...\n";
 		return 0;
 	}
 
--- a/syncoid
+++ b/syncoid
@@ -290,7 +290,10 @@ sub syncdataset {
 
 	if (!defined $sync) {
 		# zfs already printed the corresponding error
-		if ($exitcode < 2) { $exitcode = 2; }
+		if (dataset_exists($sourcehost, $sourcefs, $sourceisroot) and $exitcode < 2) {
+			$exitcode = 2;
+		}
+
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


mschout:
Ah ok.  I'll take a peek at that PR in the next few days.  Thanks.

mschout:
I changed it so that if `getzfsvalue()` is called in array context, it returns `($value, $error)`, and handle it now at the point where we fetch the `sanoid:sync` property.   Works for me.

mschout:
@phreaker0 let me know if the changes are ok, or if you would prefer to see this solved in a different way.  I've been using this branch for about a month now with no issues on both my main desktop and my laptop.

Issue #394, opened 2019-06-01
Day '0' out of range 1..30 at /usr/local/sbin/sanoid line 413


ryanjaeb:
I'm setting up sanoid for the first time today, so it's possible I've configured something wrong, but I followed the CentOS instructions and `sanoid.conf` example almost exactly.  It seems like this broke after UTC rolled over to June 1:

```bash
TZ=UTC date && TZ=UTC /usr/local/sbin/sanoid --take-snapshots --verbose
```
```
Sat Jun  1 04:20:53 UTC 2019
INFO: taking snapshots...
Day '0' out of range 1..30 at /usr/local/sbin/sanoid line 413.
```

With my local timezone (-600) which is still at May 31:

```bash
date && /usr/local/sbin/sanoid --take-snapshots --verbose
```

```
Fri May 31 22:22:48 CST 2019
INFO: taking snapshots...
taking snapshot neo/enc/images/testing@autosnap_2019-05-31_22:22:48_monthly
taking snapshot neo/enc/images/testing@autosnap_2019-05-31_22:22:48_daily
taking snapshot neo/enc/images/testing@autosnap_2019-05-31_22:22:48_hourly
taking snapshot neo/enc/images/testing@autosnap_2019-05-31_22:22:48_frequently
INFO: cache expired - updating from zfs list.
```

normanu:
Same issue here and getting 100's of mails about it

shodanshok:
I confirm the bug, which was indtroduced here: https://github.com/jimsalterjrs/sanoid/pull/362
Sorry for the inconvenience; I am thinking about a workaround, which I should publish later today.

shodanshok:
Hi, here you can find the fast fix: https://github.com/shodanshok/sanoid/commit/ada4240afa0a33bf26918f08ac394b5194b1b80c

The fix basically reverts to the method to calculate the previous day in the special case of the first day of month. Note that this can lead to some duplicate snapshot in the rare case of DST happening the first day of a month (which was the reason I authored the change in the fist place).

For manual patching, copy/save the code above and use the `patch <your_sanoid_file> -i <this_patch_file>` to apply.

```
412,413c412,417
<                                               $preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
<                                               $lastpreferred = timelocal(@preferredtime);
---
>                                               if ($preferredtime[3] > 1) {
>                                                       $preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
[...]

PR #395, opened 2019-06-01
[Regression fix] Change of DST handling

shodanshok:
As discovered today, the new DST handling approach (which I authored here https://github.com/jimsalterjrs/sanoid/pull/362) has a serious problem: the first day of a month, the variable `preferredtime[3]` become `0`, causing an exception. See here for an example of the problem: https://github.com/jimsalterjrs/sanoid/issues/394

I issued an immediate fix here: https://github.com/shodanshok/sanoid/commit/ada4240afa0a33bf26918f08ac394b5194b1b80c

However, I am re-evalating the base approach. The right solution would be to use Perl `Date::Time` class to handle date/time, but this requires an external Perl module (increasing the dependencies list)

So, in order to fix the problem without requiring new dependencies *and* in the simpler manner possible, I propose this PR. The only downside compared to current code is an additional `daily` snapshot taken at DST change.

If you want no additional snapshot, the two methods can be combined; for example:

```
412,413c412,417
<                                               $preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
<                                               $lastpreferred = timelocal(@preferredtime);
---
>                                               if ($preferredtime[3] > 1) {
>                                                       $preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
>                                                       $lastpreferred = timelocal(@preferredtime);
>                                               } else {
>                                                       $lastpreferred -= 60*60*24; # workaround for first day of month
>                                                       if ($lastpreferred > time()) {$lastpreferred -= 60*60} # skip an hour due to unhandled DST crossing
>                                               }
```

Anyway, the current patch should be reverted to avoid *no* snapshots taken on the first day of each month. Sorry for introducing the problem.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -408,14 +408,8 @@ sub take_snapshots {
 					push @preferredtime,($datestamp{'mon'}-1); # january is month 0
 					push @preferredtime,$datestamp{'year'};
 					$lastpreferred = timelocal(@preferredtime);
-					if ($lastpreferred > time()) {
-						if ($preferredtime[3] > 1) {
-							$preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
-							$lastpreferred = timelocal(@preferredtime);
-						} else {
-							$lastpreferred -= 60*60*24; # workaround for first day of month
-						}
-					}
+					if ($lastpreferred > time()) { $lastpreferred -= 60*60*24; } # preferred time is later this day - so look at yesterday's
+					if ($lastpreferred > time()) { $lastpreferred -= 60*60; } # DST crossing detected, skip another hour
 				} elsif ($type eq 'weekly')   {
 					# calculate offset in seconds for the desired weekday
 					my $offset = 0;
--- a/sanoid
+++ b/sanoid
@@ -408,10 +408,8 @@ sub take_snapshots {
 					push @preferredtime,($datestamp{'mon'}-1); # january is month 0
 					push @preferredtime,$datestamp{'year'};
 					$lastpreferred = timelocal(@preferredtime);
-					if ($lastpreferred > time()) {
-						$preferredtime[3] -= 1; # preferred time is later today - so look at yesterday's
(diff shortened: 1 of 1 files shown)


shodanshok:
@jimsalterjrs  @phreaker0  Please review the proposed PR. Thanks.

shodanshok:
@jimsalterjrs @phreaker0 any chances to review the patch? The issue seems quite important, and a fix is required. Thanks.

shodanshok:
@phreaker0 Thanks for reviewing the patch. It is true, a single duplicate daily snapshot is to be expected. It should be avoided with additional checks, but I am afraid to complicate the snapshot-decision code. I would argue that a single duplicate daily snapshot (which somewhat compensate the *missing* hourly snapshot due to DST crossing) is a very small price to pay for solving the original problem (dozens of duplicate snapshot) in the simple possible manner.
@jimsalterjrs any clue?

phreaker0:
@shodanshok @jimsalterjrs i made a pr (#406) with an alternative which solves the problem technically correct (no missing or duplicate snapshots) but by keeping the working/complex code.

shodanshok:
@phreaker0 well done! To be honest, due to my obsession with simplicity, I continue to *slightly* like more the simpler approach proposed here (which is a one-line change vs the original implementation). After all, one can always use `TZ=UTC` and avoid the missing/duplicate snapshot. However, it is very nice to have an alternative (and more correct) implementation to evaluate; thanks.
@jimsalterjrs its your choice; I'll be happy with any one of the proposed changes.

gdevenyi:
I wrote up a small proposal at #409 about using the DateTime module to handle comparing snapshots using standardized comparions and reproducible checks. Please take a look.

shodanshok:
@jimsalterjrs any thoughts on the matter? We should really fix the git master. While I like the simpler approach detailed here, I would be happy with *any* working patch. Thanks.

PR #399, opened 2019-06-06
don't use hardcoded paths

### LINKED ISSUE #393: Don't hardcode paths in sanoid
Hello, I am very interested in using Sanoid however on NixOS almost no command is in the standard location which presents a problem as commands like `ps`, `zfs`, and `zpool` are all defined as variables with a hard coded path to their assumed locations.

It would be great if it could just use `$PATH` to find them but failing that some way to override their locations is a must IMO. And not just for NixOS as I imagine some other weird distro has these in non standard paths too.s

phreaker0:
Fixes #393


Nekroze:
Can confirm this works on NixOS, Thanks @phreaker0!

sanoid
```diff
@@ -29,10 +29,13 @@ if (keys %args < 2) {
 	$args{'verbose'} = 1;
 }
 
-my $pscmd = '/bin/ps';
+# for compatibility reasons, older versions used hardcoded command paths
+$ENV{'PATH'} = $ENV{'PATH'} . ":/bin:/sbin";
```
hartzell: Two questions here:

1. Is is it the task of the application to set up $PATH, or the enclosing environment?
2. Are there really accounts/configs that don't *already* have `/bin` and `/sbin` on the path?

phreaker0:
> As part of the same change, can "someone" debride the hardcoded paths in syncoid to `pv`, `lzop`, `mbuffer` and `sudo`, which live in `/usr/local/bin` on FreeBSD systems?

already done here #270

sanoid
phreaker0: 1) no, but it's there to be sure to not break any existing installs
2) no idea :-)

jimsalterjrs: @phreaker0 has this been tested to NOT break usage in cron with its extremely restricted default PATH on, eg, Debian and Ubuntu systems?

If we break sanoid/syncoid in cron env with their brutally restricted PATH, we're going to screw a WHOLE lot of people out of their backups until they figure out what's wrong, and it's not the simplest thing to troubleshoot. Not understanding that cron PATH is much smaller than interactive PATH has been frustrating *nix users for decades.

phreaker0:
Was a long time ago :-) So I tested it again just now with completely empty PATH:

crontab:
```
PATH=' '

* *     * * *   root    TZ=UTC /usr/sbin/sanoid --cron --verbose
```

mail from cron:
```
Return-Path: <root@stan>
X-Original-To: root
Delivered-To: root@stan
From: root@stan (Cron Daemon)
To: root@stan
Subject: Cron <root@stan>    TZ=UTC /usr/sbin/sanoid --cron --verbose
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit
X-Cron-Env: <PATH=>
X-Cron-Env: <SHELL=/bin/sh>
X-Cron-Env: <HOME=/root>
X-Cron-Env: <LOGNAME=root>
Date: Thu, 16 Jan 2020 21:45:06 +0100 (CET)
Status: O
X-UID: 1

INFO: taking snapshots...
taking snapshot storage/aaa@autosnap_2020-01-16_20:45:03_frequently
taking snapshot storage/bbb@autosnap_2020-01-16_20:45:03_frequently
taking snapshot storage/ccc/ccc@autosnap_2020-01-16_20:45:03_frequently
taking snapshot storage@autosnap_2020-01-16_20:45:03_frequently
taking snapshot storage/ddd@autosnap_2020-01-16_20:45:03_frequently
taking snapshot storage/eee@autosnap_2020-01-16_20:45:03_frequently
INFO: pruning snapshots...
```

jimsalterjrs:
How is it finding the binaries with no pay and no hard code? (Sorry I'm not 
reviewing the actual code and answering my own question, replying on mobile 
from a client site.)

Please forgive any typos: tiny screens and swiping keyboards are not a 
recipe for accuracy.

phreaker0:
@jimsalterjrs to keep backwards compatibility/making sure nothing breaks, sanoid adds the necessary base paths for all commands and the end of PATH:

```
# for compatibility reasons, older versions used hardcoded command paths
$ENV{'PATH'} = $ENV{'PATH'} . ":/bin:/sbin";
```

PR #404, opened 2019-06-11
Correct yum package name for perl-Capture-Tiny

mvnural:
yum install perl-capture-tiny fails in CentOS (6 & 7) as the correct package name should be perl-Capture-Tiny. Updated it accordingly.


phreaker0:
there is already an PR for that :-) #390

mvnural:
Thanks, should've checked before submitting this.

PR #405, opened 2019-06-13
Documentation buglet, --sshkey specifies the *key*

hartzell:
The `--sshkey` option specifies the ssh key, *not* the ssh public key.  Empirically, it fails if you point it at a public key and works if you point it at the private part.  Make sense since it's being used as the arg to `-i` in the `ssh` command line.


phreaker0:
thx

PR #175, opened 2017-12-06
DST fixes

phreaker0:
sanoid uses the local time for snapshot names and planning, this patch adds correct DST handling to avoid current problems with multiple daily snapshots in the same day and a missing hourly snapshot when DST ends.
I also added a test to verify the correct behaviour of the patch.

The one thing we have to decide is how to name the duplicate hourly timestamp which is occurring when DST ends and the hour is turned back by one. The patch currently uses "_y" before the hourly snapshot type, it doesn't have a special reason.

Fixes #155

PR #406, opened 2019-06-14
revert to old dst handling and make it nice

### LINKED ISSUE #394: Day '0' out of range 1..30 at /usr/local/sbin/sanoid line 413

phreaker0:
This PR is a follow up to #175 and makes it basically complete by only suffixing the duplicate snapshot name caused by the DST hour rollback. It's an alternative to the proposed #362, #395. This one seems to work correctly (based on the long running tests and it's currently used in the last release) without any duplicate or missing snapshots but of course the dst handling leads to code which is more difficult to read and understand.

The snapshots surrounding an DST change by rolling back one hour looks like this after this patch:
 
```
...
sanoid-test-2@autosnap_2017-10-28_22:00:00_hourly
sanoid-test-2@autosnap_2017-10-28_23:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_00:00:00_daily
sanoid-test-2@autosnap_2017-10-29_00:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_01:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_02:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_02:00:00dst_hourly
sanoid-test-2@autosnap_2017-10-29_03:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_04:00:00_hourly
sanoid-test-2@autosnap_2017-10-29_05:00:00_hourly
...
```

Fixes #394 
Closes #395


jimsalterjrs:
Should I merge #395 first? Or should this be merged and #395 rejected?

gdevenyi:
I wrote up a small proposal at #409 about using the DateTime module to handle comparing snapshots using standardized comparions and reproducible checks. Please take a look.

phreaker0:
@jimsalterjrs Thats your decision to make (both @shodanshok and me are biased), either this or #395 should be merged but not both. This PR solves the problems correctly for the user (no duplicate snapshots and no lost hourly snapshots for local timezone users) in contrast to @shodanshok  solution which is simpler but an hourly snapshot is losts and there is at least one duplicate daily snapshot.

phreaker0:
@jimsalterjrs there was the idea from @gdevenyi  in #410 to just include the hour shift information to the snapshot name if local time is used. This would eliminate the complete dst handling code entirely and it would become a lot easier. Furthermore no suffixes are needed because there can't be a duplicate snapshot, as the hour shift changes on DST boundary. Would this be acceptable to you? To sum it up:

If UTC is used, nothing would change:
2019-06-19T16:43:47
But if local time is used it would look like:
2019-06-19T12:43:47-04:00

shodanshok:
@phreaker0 @gdevenyi The idea is indeed very good. However it will change the snapshot name format, which can have a non-trivial impact on the old-snapshot-trimmig code. Maybe this can be queued for a new sanoid point release, eg: 2.1.x or 3.x.x, keeping a less invasive change for 2.0.x? @jimsalterjrs any thoughts?

phreaker0:
The old snapshot trimming code should still handle the old format as if is would be in utc format, this way no snapshots are left behind and the schedule error is only up to 12/24? hours big for pruning.

skrenes:
Can we please get this or the other PR merged? It's been months with this serious problem.

shodanshok:
@phreaker0 @jimsalterjrs sorry to reopen this issue, but I just noticed that on DST->noDST crossing (which happened today in Italy), `sanoid` tries to create an already-existing snapshot, causing a ZFS error and a flurry of notification from `cron` on all affected servers. Indeed, it seems that the current code *depends* on ZFS erroring out when creating a duplicate snapshot (see here: https://github.com/jimsalterjrs/sanoid/blob/master/sanoid#L590-L591)

For example, earlier this night I received many notifications similar to that:
`cannot create snapshot 'tank/kvm@autosnap_2020-10-25_02:00:01_hourly': dataset already exists`

While nothing wrong was really done on the systems, receiving such notifications is quite annoying if you are managing many servers. Coupled with the added complexity of the current DST handling code (which even commands some code duplication, see the example linked above) I really wonder if the simpler approach described in https://github.com/jimsalterjrs/sanoid/pull/395 can be a better fit.

Any thoughts? Thanks.

gdevenyi:
#410 still

shodanshok:
@gdevenyi yeah, using perl DateTime probably is the correct approach, but it introduces additional dependencies - which I am not sure are really wanted.

PR #435, opened 2019-09-20
Update syncoid version to v2.0.2 also...

fryfrog:
Turns out there is a version in `syncoid` that was missed too. These should probably just pull from the VERSION file some how?


fryfrog:
This is obviously worth merging, but obviously *not* worth re-releasing. ;)

PR #438, opened 2019-09-25
proper changelog for the last two versions and packaging updates

### LINKED ISSUE #436: Please update the packaging to correspond with latest version numbering
Hi There,

Using the DEB build instructions it's unable to upgrade the v2.0.0 package on systems as the 

https://github.com/jimsalterjrs/sanoid/blob/master/packages/debian/changelog

The log is still on version 2.0.0, the fix would be to add new version with

dch -v 2.0.1
dch -v 2.0.2

Thanks

phreaker0:
I digged through all the pull requests and commit and wrote a nice changelog again for the last two versions. Furthermore I updated all the package control files and applied the version where it was still the old one.

Closes #435
Fixes #436


jimsalterjrs:
Thank you Christoph!

phreaker0:
Wow that was fast :-) Maybe I could point your attention to the DST problem here: https://github.com/jimsalterjrs/sanoid/pull/406#issuecomment-505501087

If you accept I would process like this:

- merge the linked PR for now as quickfix
- some time in the next weeks I will try to change the local time handling (only if utc is not used) to include the time information to the snapshot names so the problem vanishes nicely

PR #443, opened 2019-10-05
Create Docker image

rakstar:
Any interest in a Docker image? I have Docker and ZFS on pretty much all my servers so having Sanoid and Syncoid in a Docker image makes it very convenient for me to set both up on all.  Putting this out there in case someone else finds it useful.  I'm open to suggestion on how to improve this PR.  Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+docker/sanoid.conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:bionic-20190912.1
+
+LABEL maintainer="King Ramos <rkramos@yahoo.com>"
+
+RUN apt update && apt install -y \
+    curl \
+    libcapture-tiny-perl \
+    libconfig-inifiles-perl \
+    lzop \
+    mbuffer \
+    pv \
+    ssh-client \
+    zfsutils-linux \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY sanoid /usr/local/bin/
+COPY sanoid.defaults.conf /etc/sanoid/
+
+COPY syncoid /usr/local/bin/
+
+CMD [ "sanoid" ]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,46 @@
+# Sanoid Docker Image
+
+All the commands below are executed in the **docker** directory which contains the **docker-compose.yml** file.  It is assumed that you have [Docker Compose](https://docs.docker.com/compose/) already installed on your system.
+
+## Building the Image
+```bash
+docker-compose build
+```
+
+## Usage
+
+Place your **sanoid.conf** in the **docker** directory.
+
+Before the first run, SSH keys need to be generated and copied to the target server.
+
+```bash
+docker-compose run --rm sanoid ssh-keygen
+docker-compose run --rm sanoid ssh-copy-id syncoid_user@target_server
+```
+
+You are now ready to go. The default command for the container is **sanoid** so the following commands are equivalent:
+```bash
+docker-compose run --rm sanoid
+docker-compose run --rm sanoid sanoid
+```
+
+To run **syncoid**, run the following:
+```bash
+docker-compose run --rm sanoid syncoid tank/foo synncoid_user@target_server:tank/foo
+```
+
+## Tips
+
+### Allow syncoid to execute ZFS without a password on the target server
+
+On the target server:
+```bash
+sudo visudo
+```
+
+Then add the following to end of the file:
+
+```
+syncoid_user ALL=(ALL:ALL) NOPASSWD: /sbin/zfs
+```
+Reference: https://www.reddit.com/r/zfs/comments/3259gb/sanoidsyncoid_support_thread/
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,17 @@
+version: '3.3'
+
+services:
+  sanoid:
+    image: rakstar/sanoid
+    build: ../
+    privileged: true
+    devices:
+      - "/dev/zfs:/dev/zfs:rw"
+    volumes:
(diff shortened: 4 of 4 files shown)


phreaker0:
Mhm, I don't think this is a good idea:

- container is privileged 
- container can directly control the zfs pools from the host
- the container needs to have matching zfs tools for the hosts kernel zfs module to work properly

the first two points are obviously needed and bad for security but the major issue I think is the last point which in my opinion contradicts the idea of the container in the first place. 

@jimsalterjrs Therefore I don't think we should include a Dockerfile with this repo, what are your thoughts?

pirate:
I think this should be reconsidered, containers are not only for security purposes, they also make deploying and orchestrating applications across a fleet much easier.

Consider an admin who manages installing and updating all dependencies on their server via docker-compose:

```bash
/tank/www.example.com
   docker-compose.yml
   data/
       database/
           ...
       logs/
           ...
/tank/backups.example.com
   docker-compose.yml
```
`/opt/backups.example.com/docker-compose.yml`:
```yaml
version: '3'
services:
    znapsend:
        image: oetiker/znapzend:master
        command: |
			/bin/bash -c "
                znapzendzetup create \
				  SRC '7d=>1h,30d=>4h,90d=>1d' tank/www.example.com/data \
   				  DST:a '7d=>1h,30d=>4h,90d=>1d,1y=>1w,10y=>1month' root@bserv:backup/home \
                znapzend --logto=/dev/stdout"
        privileged: true
        devices:
            - /dev/zfs
```

- the entire znapsend setup is declaratively defined in one file that can be committed to version control
- the admin does not have to install packages or manage any of znapsend's child dependencies on the host os
- the admin can easily pin a specific version or update the image to the latest version by running `docker-compose pull`
- it's made explicit to the admin that this container has full privileged access to the host and the /dev/zfs device

Being able to use sanoid in this way would make it much easier to drop into existing docker-centric setups (I personally like sanoid slightly better than znapsend, and I'd love to use it in this way). It's not a compromise of Docker principles as long as the tradeoffs and purpose in this case (easier orchestration, not security) are made explicit to users.


On the topic of requiring kernel modules, it is standard for many low-level networking or filesystem containers to expect their modules to already be present on the host system, take Wireguard for example:
```yaml
services:
    wireguard:
        image: linuxserver/wireguard
        sysctls:
            - net.ipv4.conf.all.src_valid_mark=1
        cap_add:
            - NET_ADMIN
            - SYS_MODULE
        volumes:
            - /lib/modules:/lib/modules
            - ./etc/wireguard/wg0.conf:/config/wg0.conf
```
This container depends on Wireguard being already installed as a kernel module on the host, similar to how znapzend depends on ZFS already being installed on the host. Considering both ZFS and Wireguard are now part of mainline 20.04 Ubuntu Desktop, and are easily installable on other linuxes, I think allowing users who have it available to use it via Docker strictly improves its usability.

I think it's especially reasonable for an app like sanoid (designed to *manipulate* ZFS volumes) to expect ZFS itself and its headers to already be present on the machine it's going to manipulate. The container doesn't have to install ZFS itself because it's not a ZFS container, it's just an app that interacts with an already present ZFS installation, which in this case is done via dylib library calls to the host machine's ZFS install.

phreaker0:
@pirate but in contrast to filesystems and other kernel modules, zfs kernel modules and userspace application/libaries need to match, or else there will be funny error messages.

@jimsalterjrs what do you think?

pirate:
AFAIK the versions needing to match is normal. At least that's the case for Wireguard, and I'm not aware of a filesystem container that doesn't require the host kernel version to match.

deviantintegral:
I was looking at doing this in lxc, and noticed the following when installing `zfs-utils`:

```
# apt -f install
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following package was automatically installed and is no longer required:
  libfreetype6
Use 'apt autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 12 not upgraded.
1 not fully installed or removed.
After this operation, 0 B of additional disk space will be used.
Setting up zfsutils-linux (0.8.3-1ubuntu12.7) ...
modprobe: FATAL: Module zfs not found in directory /lib/modules/5.4.0-67-generic
zfs-import-scan.service is a disabled or a static unit, not starting it.
zfs-import-scan.service is a disabled or a static unit, not starting it.
Job for zfs-mount.service failed because the control process exited with error code.
See "systemctl status zfs-mount.service" and "journalctl -xe" for details.
invoke-rc.d: initscript zfs-mount, action "start" failed.
● zfs-mount.service - Mount ZFS filesystems
     Loaded: loaded (/lib/systemd/system/zfs-mount.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Sun 2021-03-21 19:47:13 UTC; 7ms ago
       Docs: man:zfs(8)
    Process: 2575 ExecStart=/sbin/zfs mount -a (code=exited, status=1/FAILURE)
   Main PID: 2575 (code=exited, status=1/FAILURE)

Mar 21 19:47:13 socks-zfs-backup systemd[1]: Starting Mount ZFS filesystems...
Mar 21 19:47:13 socks-zfs-backup zfs[2575]: Failed to initialize the libzfs library.
Mar 21 19:47:13 socks-zfs-backup systemd[1]: zfs-mount.service: Main process exited, code=exited, status=1/FAILURE
Mar 21 19:47:13 socks-zfs-backup systemd[1]: zfs-mount.service: Failed with result 'exit-code'.
Mar 21 19:47:13 socks-zfs-backup systemd[1]: Failed to start Mount ZFS filesystems.
dpkg: error processing package zfsutils-linux (--configure):
 installed zfsutils-linux package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 zfsutils-linux
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

How did the Dockerfile build without errors? It seems like the mount service would have to be masked.

PR #170, opened 2017-11-26
Debian packaging

rlaager:
I got to thinking about packaging sanoid again. I see some work has been done in that direction. Here is a series of fixes, plus a TODO file outline additional work that needs to be done.

I've been using sanoid for some time, so I'd be interested in packaging it for Debian. Are you still agreeable to that?

Note that the best practice is to have the packaging in a separate repository from upstream. I'm submitting this here so that the effort isn't wasted if someone else picks this up. But if I'm doing it, the packaging will move.

PR #446, opened 2019-10-23
Debian packaging

peterNordin:
This is a rebase of #170 
It contains the original work excluding one commit that removed sleepymutex, (based on comment in the original pull-request)


jimsalterjrs:
Thank you Peter!

Issue #420, opened 2019-07-11
Running post-sync command on newest snapshot in syncoid?


ser:
Any idea how do I run a post-sync command on the newest snapshot of each dataset synced via syncoid?

redmop:
Currently, you need to have your script find the latest snapshot. I just put in issues to help with that: #418 #419

phreaker0:
@redmop those issues refer to sanoid not syncoid
@ser currently there is no direct support for that in syncoid, you need to run a script yourself after syncoid is finished.

Issue #419, opened 2019-07-05
Pass dataset to pre/post snapshot script


redmop:
This can let the script know what datasets are going to be snapshotted, so it can react accordingly. This lets us reuse scripts instead of having a script for each dataset.

Something like this would be particularly handy:
pre_snapshot_script = /path/to/script.sh --snapshot-dataset **$sanoid-snapshot-dataset** --some-other-sexy-argument

tiedotguy:
The env vars `SANOID_TARGET` and `SANOID_SNAPNAME` are passed to pre, post, and prune snapshot scripts.  You need to be careful when using `recursive=zfs` with pre and post, as the script is only invoked for the base, not for each child.

It's not greatly documented, but it's mentioned briefly [here](https://github.com/jimsalterjrs/sanoid/blob/master/sanoid.conf#L94-L95).

redmop:
Thanks for noticing that, Do you know how snapname will handle multiple snapshots to be taken at the same time (such as freq, and hourly at the same time)

tiedotguy:
Unfortunately, they are all distinct snapshots.  If you're putting a database in to hot backup mode for instance, it will be put in and taken out multiple times.

It **looks** like they go in scale order (yearly, monthly, etc), so you could detect the first one, but I'm not 100% certain you can tell which is the last one.

redmop:
I'm just wanting to know the contents of that variable when multiple snapshots are going to be taken, and how the snapshots are delimited. I think can grep/awk my way to what I need from that.

tiedotguy:
Right - I mean it's not exposed like that.  Every snapshot, even for the same dataset, has its own pre/post script executed:

```
SANOID_TARGET=tank/dataset SANOID_SNAPNAME=autosnap_date_yearly /your/pre-script.sh
zfs snapshot tank/dataset@autosnap_date_yearly
SANOID_TARGET=tank/dataset SANOID_SNAPNAME=autosnap_date_yearly /your/post-script.sh

SANOID_TARGET=tank/dataset SANOID_SNAPNAME=autosnap_date_monthly /your/pre-script.sh
zfs snapshot tank/dataset@autosnap_date_yearly
SANOID_TARGET=tank/dataset SANOID_SNAPNAME=autosnap_date_monthly /your/post-script.sh

SANOID_TARGET=tank/dataset SANOID_SNAPNAME=autosnap_date_weekly /your/pre-script.sh
[...]

Issue #418, opened 2019-07-05
Pass snapshot type(s) to pre/post snapshot script


redmop:
This can let the script know what kind of snapshot(s) are going to be taken, so it can react accordingly.

Something like this would be particularly handy:
pre_snapshot_script = /path/to/script.sh --snapshot-type **$sanoid-snapshot-type** --some-other-sexy-argument

PR #447, opened 2019-11-02
More script information

tiedotguy:
This passes more information to pre/post/prune scripts.  It should maintain full backwards compatibility, and adds some more documentation to the README.md file.

I believe this will help closing #418, #419, and #420

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,51 @@ Which would be enough to tell sanoid to take and keep 36 hourly snapshots, 30 da
 
 	Show help message.
 
+### Sanoid script hooks
+
+There are 3 scripts which can optionally be executed at various stages in the lifecycle of a snapshot:
+
+##### `pre_snapshot_script`
+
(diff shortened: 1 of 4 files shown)


phreaker0:
@tiedotguy very nice rewrite of the snapshot options code. I like this but I think the pre/post snapshots should only be executed once for a single datasets and the scripts should get all snapshots which will be taken (daily, hourly, ...). Do you mind if I add this on top of your PR?

tiedotguy:
Not at all, go ahead.  Since this is a bit old, and I don't actually know perl (I mostly just muddled through it), I'm not fully across all the way the changes interact, the one thing I would suggest to be careful of is:

1. maintaining backwards compatibility, or
2. making it clear that this breaks backwards compatibility (which is fine to do, as long as long as it's clear)

At the time, I went with #1, but I _think_ passing all the datasets might break it because it changes the behavior of the existing `SANOID_TARGET` var?

tiedotguy:
Closing this in favor of #497

PR #463, opened 2019-12-06
Move sanoid cache and lock files to subdirectories

lopsided98:
I would like to run sanoid as an unprivileged user using ZFS privilege delegation. This works fine, except sanoid tries to write files directly into `/var/cache` and `/var/run`, which can only be written to by root.

I would like to place sanoid's files in subdirectories, ie. `/var/cache/sanoid` and `/var/run/sanoid`, which will allow me to use systemd's `CacheDirectory` and `RuntimeDirectory` options to create those folders with the correct permissions.

Rather than just changing the hardcoded directories, I decided to add `--cache-dir` and `--run-dir` options that allow the directories to be specified. I defaulted them to `/var/cache/sanoid` and `/var/run/sanoid`, but they could default to `/var/cache` and `/var/run` if there are backwards compatibility concerns I didn't think of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ Which would be enough to tell sanoid to take and keep 36 hourly snapshots, 30 da
 
 	Specify a location for the config file named sanoid.conf. Defaults to /etc/sanoid
 
++ --cache-dir
+
+	Specify a directory to store the zfs snapshot cache. Defaults to /var/cache/sanoid
+
 + --take-snapshots
 
 	This will process your sanoid.conf file, create snapshots, but it will NOT purge expired ones. (Note that snapshots taken are atomic in an individual dataset context, <i>not</i> a global context - snapshots of pool/dataset1 and pool/dataset2 will each be internally consistent and atomic, but one may be a few filesystem transactions "newer" than the other.)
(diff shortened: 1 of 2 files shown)


jimsalterjrs:
Do you see any problems here, @phreaker0 ?

Has anybody tested this in actual use yet, @lopsided98 ?

phreaker0:
No, but I didn't test it yet. But I like it.
We just need to add a note to the release page that the old files in /var/cache and /var/run can be removed as they are not longer used after this change.

redmop:
Can we try and detect files in the old location and throw up a WARN/INFO/something?

phreaker0:
I tested it and it works as expected. Merge time :-)

@redmop I'm the opinion that this is more a packaging thing (Debian postinst for example should look for the old file and delete it after upgrading).

redmop:
What about people that dont use packaging?

phreaker0:
@redmop they will need to read the release notes :-) And if the file remains unused lying around which is not nice but not critical either.

Hooloovoo:
I think this broke:
```
if (keys %args < 2) {
	$args{'cron'} = 1;
	$args{'verbose'} = 1;
}
```
as two additional keys ("cache-dir" and "run-dir") were added to %args before this is evaluated.

PR #473, opened 2019-12-28
remove invalid locks caused by race conditions or else the block the …

### LINKED ISSUE #391: sanoid_cacheupdate.lock breaking due to likely race condition
```
$ sanoid --version
/usr/sbin/sanoid version 2.0.1
(Getopt::Long::GetOptions version 2.50; Perl version 5.28.1)

$ cat /sys/module/zfs/version
0.7.12-2

$ uname -a
Linux hithaeglir 4.19.0-4-amd64 #1 SMP Debian 4.19.28-2 (2019-03-15) x86_64 GNU/Linux
```

---

I have 2 recurring crons that run on my machine that use `sanoid`:

1. The "default" systemd `timer`/`service` unit that comes with sanoid that runs `--take-snapshots` every 15 minutes
1. A zfs monitoring script that runs every ~30 seconds via `monit` that runs `--verbose --readonly --monitor-snapshots`

Note: neither of these commands use the `--force-update` option.

I've been seeing an intermittent issue (that be as long as a day or so between occurrences, to back-to-back occurrences):

```
$ monit status zfs
Monit 5.25.2 uptime: 13h 16m

Program 'zfs'
  status                       Status failed
  monitoring status            Monitored
  monitoring mode              active
  on reboot                    start
  last exit value              1
  last output                  ERROR: /var/run/sanoid_cacheupdate.lock is invalid.
  data collected               Fri, 24 May 2019 07:46:00
```

When this error appears, the `/var/run/sanoid_cacheupdate.lock` file contains

```
30842
/usr/bin/perl /usr/sbin/sanoid --take-snapshots
nly --monitor-snapshots
```

[As far as I can tell from `sanoid`'s source code, it expects 2 lines in that file](https://github.com/jasonblewis/sanoid/blob/4a3e93372c3502a7a4d9e2202d147ab00fd23559/sanoid#L942-L948): the first, I'm guessing, is the PID of the process holding the lock; the second is the full command of the process holding the lock.

Between run of either of these crons, I can see that the lockfile should _normally_ be non-present. I imagine it's only created when sanoid is updating its cache and then it's removed. The problem is that one of these processes is leaving the lockfile around after it's done, and in a bad state (note the broken third line).

I'm guessing that, from time to time, these two crons are running at exactly the same moment and simultaneously trying to write the lockfile. One of them fails, and thus leaves the broken lockfile in place. Deleting the lockfile allows both crons to run until the issue occurs again.

---

I'm not really sure what could be done on `sanoid`'s side to fix this. Maybe some kind of timeout and retry? If the race condition is occurring outside of the `sanoid` process itself... ¯\_(ツ)_/¯

In the meantime, I think I've mitigated the issue on my system by using `flock -w <secs> /tmp/sanoid.lock` on both crons. That way, if any automatic sanoid process is running, the another will wait for it to finish before doing its thing.

However, this is mildly clunky, and if there's a way for `sanoid` to handle this internally, that'd be preferred. If nothing can be done (which is understandable), maybe a note can be added to the README or manpage indicating the issue and the possible solution? If that's the route we end up going, I'm happy to write up a PR for it.

---

Anyway, this is a fantastic tool otherwise and it's been incredibly useful for keeping my ZFS-based system(s) automated. Thanks for writing this and keep up the good work 👍

### LINKED ISSUE #468: /var/run/sanoid_pruning.lock invalid
I've been trying to isolate an issue with sanoid cron job for some months now to no avail.

Occasionally the sanoid_pruning.lock file will become corrupted including some code in addition to the pid number:
```
# cat /var/run/sanoid_pruning.lock
23087
/usr/bin/perl /usr/sbin/sanoid --cron
snapshots --verbose
```
This does not happen all the time, but then it does, I start getting error mails once each cron run (1 per minute).
I have to login to the server and manually remove the bad lock file.
The original process that created the lock file is not executing any more and nobody deleted the file.

This is sanoid 2.0.3 "compiled" from git and installed via deb package on debian 10.1

Everything is up to date. The zfs pool I'm trying to snapshot/prune and remote replicate with syncoid is a 10 drive mirrored in 5x2 mirrors.

What can I do next?

phreaker0:
…critical function of snapshot taking until manual intervention

Fixes #468
Fixes #391

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -1389,7 +1389,9 @@ sub checklock {
 	# make sure lockfile contains something
 	if ( -z $lockfile) {
 	        # zero size lockfile, something is wrong
-	        die "ERROR: something is wrong! $lockfile is empty\n";
+	        warn "WARN: deleting invalid/empty $lockfile\n";
+	        unlink $lockfile;
+	        return 1
 	}
 
 	# lockfile exists. read pid and mutex from it. see if it's our pid.  if not, see if
@@ -1400,7 +1402,9 @@ sub checklock {
 	close FH;
 	# if we didn't get exactly 2 items from the lock file there is a problem
 	if (scalar(@lock) != 2) {
-	    die "ERROR: $lockfile is invalid.\n"
+	    warn "WARN: deleting invalid $lockfile\n"
+	    unlink $lockfile;
+	    return 1
 	}
 
 	my $lockmutex = pop(@lock);


sanoid
```diff
@@ -1400,7 +1402,9 @@ sub checklock {
 	close FH;
 	# if we didn't get exactly 2 items from the lock file there is a problem
 	if (scalar(@lock) != 2) {
-	    die "ERROR: $lockfile is invalid.\n"
+	    warn "WARN: deleting invalid $lockfile\n"
```
zeigerpuppy: missing end `;`

zeigerpuppy:
I tried compiling and running this patch (on a fresh pull of the `master` branch) and `sanoid` fails to run with error:

```
syntax error at /usr/sbin/sanoid line 1406, near "unlink"
Execution of /usr/sbin/sanoid aborted due to compilation errors.
```

I found the issue (missing `;` on line 1405)

On first run, the newly compiled `sanoid` throws error:
```
WARN: deleting invalid/empty /var/run/sanoid_cacheupdate.lock
WARN: deleting invalid/empty /var/run/sanoid_cacheupdate.lock
ERROR: No valid lockfile found - Did a rogue process or user update or delete it?
```

On second run, it runs about a hundred snapshots and then throws:

```
WARN: deleting invalid/empty /var/run/sanoid_cacheupdate.lock
ERROR: No valid lockfile found - Did a rogue process or user update or delete it?
```

On third run, it does some pruning, then errors again

```
WARN: deleting invalid/empty /var/run/sanoid_pruning.lock
ERROR: No valid lockfile found - Did a rogue process or user update or delete it?
```

the next set of runs, seems to error out after each dataset is pruned (with the same error as above)

so it does look like progress but not quite getting over the missing/invalid lockfile

zeigerpuppy:
if I keep running `sanoid`, it does one data set prune at a time.  As it can't keep track of the cache, it end up trying to prune data sets repeatedly, throwing lots of these sort of errors:

```
INFO: pruning zpool2/images/ldap1@autosnap_2020-01-14_18:40:01_hourly ...
could not find any snapshots to destroy; check snapshot names.
could not remove zpool2/images/ldap1@autosnap_2020-01-14_18:40:01_hourly : 256 at /usr/sbin/sanoid line 327.
```

persisting, with executing `sanoid`, it eventually refreshes the cache (I think when the next tick of my `frequent_period = 15` ticks over):

```
WARN: deleting invalid/empty /var/run/sanoid_cacheupdate.lock
ERROR: No valid lockfile found - Did a rogue process or user update or delete it?
```

I thought I'd just keep going to see if it helps...
but the errors persist.

At least it's managed to very manually prune now :)

sanoid
phreaker0: thx

zeigerpuppy:
I tracked down the cause of the error in my case.
It's a bit embarrassing... `/run` had run out of free space due to a runaway process.
I think this must have prevented sanoid from updating its cache.

I should have noticed it earlier...

gdevenyi:
> It's a bit embarrassing... /run had run out of free space due to a runaway process.

Maybe sanoid should be better propagating up the error it encountered so its more obvious whats going wrong?

PR #494, opened 2020-02-09
Use grep rather than -p with ps for busybox

lightray22:
For those using Alpine Linux or other busybox-based Linuxes, -p is not an option for ps.  Using grep instead seems to work fine.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -1428,18 +1428,28 @@ sub checklock {
 		# we own the lockfile. no need to check any further.
 		return 2;
 	}
-	open PL, "$pscmd -o args= | grep $lockpid |";
+
+	open PL, "$pscmd -o pid,args | grep \"$lockpid \" | grep -v \"grep $lockpid\" |";
 	my @processlist = <PL>;
 	close PL;
 
-	my $checkmutex = pop(@processlist);
-	chomp $checkmutex;
+	if (scalar @processlist > 0)
+	{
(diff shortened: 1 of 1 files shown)


phreaker0:
@lightray22 I just checked busybox ps (`busybox ps -o args=`) with the same behaviour, no pids in the output. I'm running this on an ubuntu system and at least there the lock check will never work. Can you verify the output in your system?

lightray22:
I'm running Alpine Linux 3.11 with Busybox.
`ps --help` gives the following output...

`BusyBox v1.31.1 () multi-call binary.
Usage: ps [-o COL1,COL2=HEADER]
Show list of processes
	-o COL1,COL2=HEADER	Select columns for display`

Just `ps` shows ...
`PID   USER     TIME  COMMAND
    1 root      0:00 /sbin/init
    2 root      0:00 [kthreadd]
    3 root      0:00 [rcu_gp]
    4 root      0:00 [rcu_par_gp]`
and so on.  

You're right that if you don't select the PID in -o, you don't get PIDs.  Missed that part.  Regardless, -p does not work in busybox ps.  It looks like it is in the Posix spec so I guess busybox is lacking here.

lightray22:
These changes are a bit ugly but seem to do the trick.  I have never written or even looked at perl before but figured I'd give fixing this a shot rather than just filing a bug.

PR #495, opened 2020-02-11
syncoid: add option for particular ssh_config file to ssh

o-m-d:
Add option --sshconfig to syncoid for passing path to particular ssh_config(5) file to ssh.


phreaker0:
added via #636

PR #510, opened 2020-02-14
catch another case in which the resume state isn't availabe anymore

phreaker0:


samburney:
This patch worked great for me.

```
Resuming interrupted zfs send/receive from rpool/volumes/vm-109-disk-0 to raidpool1/backup/syncoid/svr1.dc1/rpool/volumes/vm-109-disk-0 (~ UNKNOWN remaining):
cannot resume send: incremental source 0x10819c471eb089af no longer exists
cannot receive: failed to read from stream
WARN: resetting partially receive state because the snapshot source no longer exists
```

PR #513, opened 2020-02-21
implemented option for direct connection via socat and busybox nc

### LINKED ISSUE #371: Request: NC support
Hey there,

Any plans to add netcat support in the near future? SSH is really slow when it comes to large datasets over 10/40gbe networks.

Thanks!

phreaker0:
This implements the desired feature to bypass ssh for sending the replication data and use a plain TCP connection. Added warnings of course that this option should not be used lightweight, the parameter option alone should be a big hint to the user :-)

An example use case: Two servers connected via a common network and via a dedicated link.
`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444 local_pool root@backup:remote_pool`

192.168.32.2 is the network for the direct link and the target host ip address. So all the unencrypted data is transferred via the dedicated link which is trusted.

The option can also be used in the case of natted network topolgies by specifying a differen listen address:

`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444,10.0.2.4:3333 local_pool root@backup:remote_pool`

Why did I use socat and busybox nc? Because it made it possible to make it really easy and clean to implement.

socat supports connection retrying which is needed because the listening socket isn't available immediately.
And the busybox netcat implementation is the only one I found which can timeout on an listening socket which is needed to abort if the connection doesn't work (firewall, argument error, ...)

Fixes #371


secabeen:
This is great!  Totally saves wasted CPU cycles on two systems connected by a trusted network doing pointless SSH encrypt/decrypt (especially when sending raw encrypted streams).

devZer0:
i like this approach (thanks for commenting my FR), but i'm not very comfortable with additional dependency on 2 tools which basically do the same and which do something,  mbuffer also can do ( as shown here:  https://github.com/jimsalterjrs/sanoid/issues/371) 

why not using mbuffer completely and do something like "timeout 10 mbuffer:LISTENPORT" on the receiving side?

devZer0:
sorry, but how can i install busybox netcat in a clean way on centos 7 (via rpm command) ?

devZer0:
do we really really need busybox nc and socat ?

couldn't we simply use:
# mbuffer -W 10 -I 8888
mbuffer: error: watchdog timeout: input stalled; sending SIGINT

and wrap "mbuffer -O host:port" tries/retries in the syncoid perl-script ?

jimsalterjrs:
@phreaker0 does this start the netcat listener on the remote side and close it again as needed, or just expect to find an always-on listener?

phreaker0:
@devZer0 Nice, It didn't read/found out that mbuffer timeout also works for non established connection, for all the other tools I tested it didn't (normal netcat, socat, ...). So I guess I can exchange busbox netcat for mbuffer then (will test this latter). But I still need socat for the connection retry options.
Doing the retry stuff in perl (as I first planned) would be way more difficult, as I can't use the existing single ssh pipe call for send/recv as syncoid tools, so I would need to rewrite much code of syncoid and it will be more error prone, as I have to start the server on the target and client on the source separately and also need to monitor them somehow, but without threads this will be difficult.
And the code would be much harder to maintain and likely wouldn't get merge.


@jimsalterjrs it will start the netcat listener on the remote side as needed and will close it again after the replication or on error

jimsalterjrs:
If you're going to reimplement with mbuffer I'll wait to test.

Is there any value to allowing either mbuffer or netcat as transports, 
keeping some of your existing work for netcat--or should we do mbuffer 
only, to minimize maintenance complexity down the road?

devZer0:
i had some little conversation with socat author/maintainer and asked for listen timeout feature in socat and convinced him it could be useful. he sent a patch with 1 day :)

as it will need some time that such enhancement will find it's way in major distros, i think there could be 2 ways to proceed:

1. use socat on the sending and mbuffer on the receiving side
2. use socat on both sides if socat is >=versionnr. with that new feature

@phreaker0 , if you like testing the socat patch i can forward it to you

furthermore, i'm feeling uncomfortable that there is a listener on the receiving side which accepts connection from everywhere for the timeframe of transfer. 

if socat is used on the receiving side, there could be easily added some security option to restrict access ( see "RANGE option group" in https://linux.die.net/man/1/socat )

devZer0:
btw, i get the following warning:

Use of uninitialized value $sourcehost in string ne at ./syncoid line 128.

TheLQ:
I get a "Use of uninitialized value $sourcehost in string ne at ./syncoid line 128." warning when using 

`./syncoid --create-bookmark -r --compress=none --insecure-direct-connection host2:4343 big10 root@host2:big8/big10`

`--insecure-direct-connection` should be able to take just a port and pull the host from the destination argument. Makes LAN backups simpler. Ideally it could automatically pick a free port but that seems to be non-trivial with `busybox nc`

phreaker0:
@TheLQ warnings are fixed, command checks are in place
@devZer0 i'm now using mbuffer for the listening socket instead of busybox nc but's it nice that socat will have a listen timeout in the future as well. The reason why mbuffer didn't work for me at first was the order of arguments, if one uses mbuffer -I 8888 -W 10 it will not timout if there is no connection, only if -W is before the -I flag. This is documented in the manpage.

@jimsalterjrs I don't see a point in supporting busybox nc as well if mbuffer can do the job as well, you can test now

phreaker0:
mhm, it doesn't work on my servers (i only tested with local addresses on my machine), need to investigate

phreaker0:
So, mbuffer behaves much differently than the other listening tools. The address provided to mbuffer isn't used as listening address but as src address whitelist. mbuffer will listen on all network interfaces.

Therefore I switched back to busybox nc as default and added an option for switching to mbuffer (in which case the specified listen address is used as an IP filter).

I also increased the default timeout to 60 seconds and made it configurable, for some of my datasets with tiny files on rust and lot's of metadata changes zfs send can be so slow that a timeout is triggered.

Command check for busybox nc and mbuffer is done according to the provided options.

examples:

**busybox nc, target and listen ip is the same (no NAT)**
`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444 local_pool root@backup:remote_pool`

**busybox nc, target and listen ip is different -> NAT**
`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444,10.0.2.4:3333 local_pool root@backup:remote_pool`

**busybox nc, target and listen ip is the same (no NAT) and timeout of 120 seconds**
`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444,192.168.32.2:4444,120 local_pool root@backup:remote_pool`

**mbuffer tcp (192.168.32.1 is src address), target and listen ip is the same and timeout of 120 seconds**
`syncoid --compress=none --insecure-direct-connection=192.168.32.2:4444,192.168.32.1:4444,120,mbuffer local_pool root@backup:remote_pool`

asche77:
Would love to see this for LAN syncs ...

geudrik:
This looks awesome, looking forward to seeing it merged

jim-perkins:
just wanted to comment that I have tested the insecure connection and it 'works for me'  using 10Gb SPF+ I was limited to about 200MB per second network transfer with ssh cipher 'aes128-ctr' . only about 150MB with the default ssh cipher.  using the insecure connection was able to sustain almost 400MB per second.  The data was being transferred between ssd's on both ends.

BTW I just copied the syncoid from   https://github.com/phreaker0/sanoid/blob/direct-connection/syncoid and dropped it into /usr/local/bin

and here is the command line tested

    /usr/local/bin/syncoid  --recursive --no-sync-snap --compress=none --insecure-direct-connection=192.168.1.30:4444,192.168.1.20:4444,120,mbuffer server_ssd root@serverx:serverx_ssd/backup

TheLQ:
I'm using this and it works fine for me. Merged with latest master and fixed the minor conflict from a new option being added, still no problems.

devZer0:
regarding https://github.com/jimsalterjrs/sanoid/pull/513#issuecomment-596450652  , socat since version 1.7.4.0 now supports option "accept-timeout" , which make it also suitable for sanoid/syncoid

http://www.dest-unreach.org/socat/doc/socat.html#OPTION_ACCEPT_TIMEOUT

accept-timeout=<seconds>
    End waiting for a connection after <seconds> [timeval] with error status. 
    
http://www.dest-unreach.org/socat/doc/CHANGES

	New option accept-timeout (listen-timeout)
	Test: ACCEPTTIMEOUT
	Proposed by Roland

tinsami1:
What else is needed for this to be included in the next release?  Or, at least, merged to master?

phreaker0:
@mailinglists35 checking your output ncat  is exiting immediately:
'Ncat: Could not resolve hostname "192.168.70.11:12345": Name or service not known. QUITTING.'

and socat retries 10 times with 1 second intervals and gives up.

mailinglists35:
oh so the $directtimeout is for socat, not for nc? it seems to be used both by nc and socat, though.

mailinglists35:
oh, sorry, nmap-ncat does not like -l IP:PORT :)

mailinglists35:
@phreaker0 I see you have a $directmbuffer hardcoded, is it usable if I switch to 1, and how? will that bypass nc?

mailinglists35:
ok, I modified my local copy of syncoid to understand nmap-ncat, since there is no busybox in EL9 repos...

MrRinkana:
For reference for people looking, and since im not sure where to put it but:
(if you are getting `nc: timed out` or `mbuffer: error: watchdog timeout: input stalled; sending SIGINT`)

**socat seems to for some reason ignore your ip routes** so if you have like me set up a (example: wireguard) gateway on your lan which port-forwards to the target at the other end of the tunnel you cannot specify the "ip routed ip", **but must use the ip of the gateway directly**.

<details>
<summary>Example:</summary>

LAN at source is: 168.192.1.1/24
target is: 10.10.10.2
gateway is: 168.192.1.2 on LAN and 10.10.10.1 in tunnel/other interface
ip route is: 10.10.10.2 via 168.192.1.2 ... ...

then the commands 
`--insecure-direct-connection=10.10.10.2:4444,..blabla... ...` **will fail**. 
`--insecure-direct-connection=168.192.1.2:4444,..blabla... ...` **will succeed!**

</details>

Took me a good while to stumble into 😅

griznog:
Tried using this today and after a bit of head scratching finally got it to work by modifying `syncoid` to build a `busybox nc` command that looked like:

```
busybox nc -l  -p 9043 -w 60
```

Could not get it to work with the arguments like it was producing with "HOST:PORT". This is on Rocky 8, so maybe I have an older/different version of `busybox`.

PR #529, opened 2020-04-20
use sudo if neccessary for checking pool capabilities regarding resum…

### LINKED ISSUE #527: zfs resume check fails for Debian target
Hi, first thanks for making sanoid, it's been really useful for me!

I noticed that `syncoid` is complaining about resume feature not available on the target system, whereas I confirmed manually that it is supported. I did some digging, and it's because of this line:

```perl
my $resumechkcmd = "$zpoolcmd get -o value -H feature\@extensible_dataset";
```

Here the check command runs as non-sudo, and calls `zpool` without path. On Debian, `zpool` is not in the `PATH` of non-root users -- hence the failure.

phreaker0:
…eable send

Fixes #527

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1067,11 +1067,11 @@ sub checkcommands {
 		my $resumechkcmd = "$zpoolcmd get -o value -H feature\@extensible_dataset";
 
 		if ($debug) { print "DEBUG: checking availability of zfs resume feature on source...\n"; }
-		$avail{'sourceresume'} = system("$sourcessh $resumechkcmd $srcpool 2>/dev/null | grep '\\(active\\|enabled\\)' >/dev/null 2>&1");
+		$avail{'sourceresume'} = system("$sourcessh $sourcesudocmd $resumechkcmd $srcpool 2>/dev/null | grep '\\(active\\|enabled\\)' >/dev/null 2>&1");
 		$avail{'sourceresume'} = $avail{'sourceresume'} == 0 ? 1 : 0;
 
 		if ($debug) { print "DEBUG: checking availability of zfs resume feature on target...\n"; }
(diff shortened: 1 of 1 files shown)


ein-stein-chen:
Wouldn't it be enough to just add `/usr/sbin` to the path? 
Checking the zpool capabilities with sudo breaks if running syncoid with an unprivileged user.

phreaker0:
@ein-stein-chen If you are running syncoid with an unprivileged user you would use the "--no-privilege-elevation" in which case the zpool command runs without sudo.

ein-stein-chen:
It would try to run zpool, but this will not succed, as zpool is in `/usr/sbin` on a current debian system.

Syncoid already adds `/bin:/usr/bin:/sbin` to the Path for compatibility reasons. ([Syncoid line 80](https://github.com/jimsalterjrs/sanoid/blame/master/syncoid#L80)) and checking for the resumable capability doesn't require any advanced privileges.

As far as I understand #527 the problem is: 
- Syncoid doesn't look in `/usr/sbin` for the  zpool executable

and not: 
- Syncoid needs to run zpool as root to check if the zpool supports resuming.

phreaker0:
@ein-stein-chen for me #527 was about the normal use cause with privilege elevation, in which case the root PATH wasn't used because of the missing sudo command for zpool which is a bug.
The compatibility path fixups are there to not break existing configuration from the time where absolute paths where used. It shouldn't add any other paths.
If you want an unprivileged user for the replication you need to make sure that this user can access all the relevant tools, in this case you would add the /usr/sbin path (or make a symlink) to the PATH env variable for the replication user.

ein-stein-chen:
thx for the explanation, maybe it would be good to add something like ` You have to ensure that syncoid finds every binary in the path.` to the help output of `--no-privilege-elevation`

PR #536, opened 2020-05-01
implemented option for keeping sync snaps

### LINKED ISSUE #533: How to solve keeping multiple snapshots on the target and not accidentally deleting the last 'common' snapshot?
I have 24 boxes that all use sanoid for snapshots and syncoid to back snapshots up to a set of remote boxes.

I goofed up a few days ago.
All the source boxes take yearly, monthly, weekly, daily, hourly, and frequent (15 minute) snapshots.
I run with the ```--no-stream``` option because I don't need every single intermediate snapshot backed up off-site.  Just a once-per-day off-site backup.  The difference between an intermediary send in 24 hours is several hundred gigs due to temporary data churn throughout the day, whereas a diff between two snapshots 24-hours apart is usually under 5 GB.

Running syncoid without ```--no-sync-snap``` creates a snapshot and sends it off-site.  But it also deletes the previous snapshot for some reason.  This means I *only* get the most current backup off-site.  I'd like to have several days worth of off-site backups to roll back to.

So I figured I would use ```--no-sync-snap```.  This mostly accomplishes what I want--it finds the latest snapshot and sends it off-site.

Unfortunately, what I failed to account for is that it's usually syncing a 'frequent' snapshot.

I had 'frequent' snapshots set to keep 96 of them hanging around.  96 snapshots / 24 hours = 4/hr (every 15 minutes).

I'm sure you can see where this is going... ;)

Due to some maintenance, backups were unable to run on-time a few days ago.  The source boxes dutifully kept removing frequent snapshots...and now there is no common snapshot between the source and target boxes.

Oops.  I immediately increased the 'frequent' snapshots to keep 4 days worth, and I'm planning to re-sync about 40 TB of data across some *really* slow Comcast connections.

Anyways....I'm trying to find a better solution to this, but reading through the docs hasn't turned up anything fruitful.

Ideally, I would love to NOT use ```---no-sync-snap``` *and* have the ability to skip deleting the previous synced snapshot and allow me to manually manage (or use a tool) to clean up old off-site backups.  Or maybe an option to ```--keep 5``` or something.

Or with ```--no-sync-snap``` tell syncoid to consider only snapshots with 'daily' in the name?  (I've seen a few tickets about a feature request).

Any pointers on how I can get data retention on the target while also not using the 'more ephemeral' snapshots?

phreaker0:
Fixes #533

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 + --source-bwlimit <limit t|g|m|k>
 
-	This is the bandwidth limit in bytes (kbytes, mbytes, etc) per second imposed upon the source. This is mainly used if the target does not have mbuffer installed, but bandwidth limits are desired. 
+	This is the bandwidth limit in bytes (kbytes, mbytes, etc) per second imposed upon the source. This is mainly used if the target does not have mbuffer installed, but bandwidth limits are desired.
 
 + --target-bw-limit <limit t|g|m|k>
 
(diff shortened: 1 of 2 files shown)


discostur:
@phreaker0 when will this PR get merged?

Issue #409, opened 2019-06-14
Proposal: Switch date and time handling to use CPAN DateTime modules


gdevenyi:
It looks like based on #406 and #395 and maybe others that the time/date handling for snapshot naming has become a bit of a problem.

Perl offers a particularly useful module DateTime, for both handling dates and times, generating standard formatting, and doing math with these objects in a sensible and standards-compliant way. It also supports sorting and comparisons between DateTime objects in perl compatible ways for tests.

Short example, get the current time in ISO UTC format as a string, then re-parse that string back into a DateTime object and extract a bunch of sub-date details
```pl
use DateTime;
use DateTime::Format::ISO8601;

$mydatetime = DateTime->now()->iso8601();

print("$mydatetime \n");

$dt = DateTime::Format::ISO8601->parse_datetime( $mydatetime );

use DateTime;
use DateTime::Format::ISO8601;

$mydatetime = DateTime->now()->iso8601();

print("$mydatetime \n");

$dt = DateTime::Format::ISO8601->parse_datetime( $mydatetime );

$year   = $dt->year;
$month  = $dt->month;          # 1-12
 
$day    = $dt->day;            # 1-31
 
$dow    = $dt->day_of_week;    # 1-7 (Monday is 1)
 
$hour   = $dt->hour;           # 0-23
$minute = $dt->minute;         # 0-59
 
$second = $dt->second;         # 0-61 (leap seconds!)
 
$doy    = $dt->day_of_year;    # 1-366 (leap years)
 
$doq    = $dt->day_of_quarter; # 1..
 
$qtr    = $dt->quarter;        # 1-4

print("$year $month $day $dow $hour $minute $second $doy $doq $qtr\n");
```

What's the exact difference in time between now and ~ a year from now.
```
use DateTime;
use DateTime::Duration;

$today = DateTime->now();
$later = DateTime->new(
    year      => 2020,
    month     => 06,
    day       => 26,
    hour      => 1,
    minute    => 30,
    second    => 0,
    time_zone => 'UTC',
);

$dur = $later->subtract_datetime_absolute($today);
print($dur->in_units( 'seconds'));
```

gdevenyi:
https://metacpan.org/pod/DateTime
https://github.com/houseabsolute/DateTime.pm/wiki

phreaker0:
@gdevenyi do you intend to port the entire code? The hard part will be detection the DST crossover and handle the duplicate snapshot accordingly.

[...]

PR #537, opened 2020-05-03
sync snap timestamp format

delxg:
Change the sync snap timestamp format to `YYYY-MM-DD_HH:MM:SS`.

Previously it was `YYYY-MM-DD:HH:MM:SS`. Note the confusing `:` between the date and the time :)


gdevenyi:
#409 strikes again

PR #556, opened 2020-06-10
RFC: Add --skip-redundant-sync-snap

abrasive:
Commonly we want to sync all files on a filesystem, even newer than the
last snapshot. So by default syncoid creates a sync snapshot and sends
that over. This can take many seconds - even if no data has changed.

This patch adds a new flag, `--skip-redundant-sync-snap`, which checks the
`written@...` property on the filesystem to see if anything has been
written since the latest snapshot. If not, it avoids creating one. This
avoids transferring anything at all if nothing has changed.

If no sync snapshot was created, pruning of old sync snapshots on the
target is also skipped.

At the moment, it uses the last snapshot (of any kind). It is assumed that snapshots are never deleted on the target, except via the syncoid pruning mechanism.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -23,7 +23,7 @@ my $pvoptions = "-p -t -e -r -b";
 my %args = ('sshkey' => '', 'sshport' => '', 'sshcipher' => '', 'sshoption' => [], 'target-bwlimit' => '', 'source-bwlimit' => '');
 GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r", "sendoptions=s", "recvoptions=s",
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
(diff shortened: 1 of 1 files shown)


abrasive:
I suspect this should make sure that the most recent snapshot is in fact a sync snapshot, so that the source and target don't lose their most recent common snap if other snaps are deleted on the source.

Or, in another embodiment, it might just find the most recent sync snapshot and check `written@...` against that snap. This would then not transfer any more recent, but also empty, non-sync snaps. For my use case that would be perfect, but I'm sure it wouldn't suit everyone.

phreaker0:
@jimsalterjrs what do you think? imho it isn't worth the effort as snapshots are very cheap.

phreaker0:
@abrasive i don't think that we will add this, with the implemented of block cloning this may also be not working as expected. I don't know i a blocked clone file would be shown in the written property.

But you can add this feature anyway with an extra script which decides if a snapshot should be taken: https://github.com/jimsalterjrs/sanoid#pre_snapshot_script

Issue #42, opened 2016-06-24
syncoid syncs all snapshots, not just the ones created by sanoid


lckarssen:
I'm not sure if this is a bug or a feature, but I noticed that syncoid syncs all snapshots between the source and destination. I expected it to only sync the ones created by sanoid (those whose names start with `autosnap`).

The reason why I'm reporting this is that I also have snapshots (not created by sanoid) that don't need to be replicated (e.g. snapshots that I take every 15 minutes).

I guess the simplest solution would be to do some grep-ing in the `getsnaps()` function, but maybe an option like `--only-autosnaps` would be more appropriate. Or, even more versatile: add an option to allow the user to select which snapshots to sync (e.g. something like: `--include-snap-type=weekly` that would only send snapshots named `autosnap_*_weekly`)

jimsalterjrs:
It's a feature, but I admit I'd like to make it an *optional* feature 
eventually.

At the moment Syncoid uses the -I argument to replicate the full snapshot 
stream natively, which captures everything in a single ZFS operation. Doing 
what you're proposing requires performing the replication in multiple steps 
- an -i from @1 to @2, then from @2 to @3, then from @3 to @4, and so forth.

That *will* happen sometime fairly soon, though, since if like to be able 
to skip certain snapshots by pattern myself. It'll just be more complicated 
than it might have seemed at first blush, and I'm still deciding how I want 
to handle the progress meter properly.

----------

(Sent from my phone - please blame any weird errors on autocorrect)

lckarssen:
Indeed, you are right, it is more complicated that I thought at first (and I didn't think of the progress bar at all :-)). 

In the mean time, I'll continue my test run of sanoid/syncoid. Thanks for both tools, so far they work well for my use case.

mailinglists35:
> since if like to be able to skip certain snapshots by pattern myself.

I was going to file a feature request for this but searching the existing issues first revealed the discussion above. I don't mind sending more volatile snapshots as part of the whole send as long as the reference syncing snapshot is one of more long term type.

I'm thinking as a quick personal solution to hardcode syncoid to exclude my `zfs-auto-snap_hourly` and `zfs-auto-snap_frequently` from the list of candidates for `$oldestsnap` or `$newsyncsnap` (I prefer `zfs-auto-snapshot` for snapshot management)

I wish it was possible as an argument on command line to exclude sync candidates by a regex filter.

[...]

PR #567, opened 2020-07-18
add --accept-snap filter to syncoid

NiklasGollenstede:
This adds a flag `--accept-snap=REGEX` to syncoid, allowing to filter the source snaps that are candidates to become the from or to sync snaps.

Syncoids default behavior to avoid the situation that the snap used on the last sync between to hosts is missing on wither end on the next sync is to create dedicated sync snaps, and prune these itself.

While I assume that that works well in most situations, it does require Syncoid to have the ZFS `destroy` permission on the target. For push replication from a production server, I'd like to have that server have as little access on the backup server as possible, especially I'd like it not to be able to mess with the existing backups.

With `--no-sync-snap`, that seems to be working quite nicely if the target user has the permissions `create,receive,mount,canmount,userprop` on the target filesystem. Except that Syncoid will likely use one of Sanoid's `_frquent` snaps and things break soon thereafter.

The new `--accept-snap` option allows to restrict the possible selection of sync snaps to those that are expected to still exist on the next sync. Obvious downside is that all changes since that pore persistent snap would not be synced.

In my case, I run this as:
```bash
syncoid \
    --recursive --no-sync-snap --create-bookmark \
    --accept-snap=^autosnap_[0-9:_-]+_(?:daily|weekly|monthly|yearly)$ \
    --no-rollback --no-privilege-elevation --sshkey /root/.ssh/syncoid_ed25519 \
    --sendoptions='w' --recvoptions='u o canmount=off' \
    ${dataset} zfs-from-${source}@${target}:backup/${source}/${dataset}
```
And it behaves as intended. I also ran the `test/syncoid` tests, and they passed.

This was suggested a few years ago in #42, but never commented on.

---

Notes:
* This repo was really easy to clone and install, thanks for that! Only thing is that it could mention to use `--reinstall` with `apt install ../sanoid_*_all.deb` when working on `master`.
* I never wrote anything in Perl before (except for PCREs), so please check that the code actually does what one would intuitively think it does.
* At least `sanoid/tests/syncoid/6_reset_resume_state2` seems to be unstable. It failed on my first execution, then passed without changes. I suspect that two snapshots were to be named the same (i.e. same timestamp).

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,11 +208,15 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 + --no-sync-snap
 
-	This argument tells syncoid to restrict itself to existing snapshots, instead of creating a semi-ephemeral syncoid snapshot at execution time. Especially useful in multi-target (A->B, A->C) replication schemes, where you might otherwise accumulate a large number of foreign syncoid snapshots.
+	This argument tells syncoid to restrict itself to existing snapshots, instead of creating a semi-ephemeral syncoid snapshot at execution time. Especially useful in multi-target (A->B, A->C) replication schemes, where you might otherwise accumulate a large number of foreign syncoid snapshots. Also useful to avoid having to give the receiving user (if not root, see --no-privilege-elevation) the destroy permission.
 
 + --create-bookmark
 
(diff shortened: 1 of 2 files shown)


redmop:
Can you stick your example in the readme?

NiklasGollenstede:
Gladly. I already stared documenting my setup for myself, but without any real plan to publish it.

So after writing a very short summary of its core idea at the end of the readme, I used this as an excuse to spend the extra time to polish my guide and add a link to its GIST to the bottom of the example.

I am not sure if this was what you intended, any comments welcome!

phreaker0:
implemented with #699

Issue #574, opened 2020-08-03
update Readme.md to reflect Ubuntu/Debian? stores sanoid.defaults.conf at /usr/share/sanoid/sanoid.defaults.conf


jim-perkins:
Thanks for all the work on the program.  Been reading up and decided to install it tonight.

Not really sanoid issue but had to run this down.  Ubuntu 20.04 provides  sanoid 2.0.3-2 .  Maybe the -2 stands for move the sanoid.defaults.conf file.  Ubuntu is storing it at /usr/share/sanoid/sanoid.defaults.conf.  As I was following along with a fresh install using Readme.md it took some time to realize Ubuntu had changed things up and the file was in place just not where it was documented.  I have no idea what is *right_way*.  Just wanted to comment that it is different.

~~Looks like~~ this package was sourced from Debian so the same ~~may apply~~ applies there.
https://launchpad.net/ubuntu/+source/sanoid/2.0.3-2 

/debian/patches/defaults-in-usr-share.patch

````
The defaults file is not a config file, store it in '/usr/share/sanoid'.

--- a/sanoid
+++ b/sanoid
@@ -36,7 +36,7 @@
 my $zpool = '/sbin/zpool';
 
 my $conf_file = "$args{'configdir'}/sanoid.conf";
-my $default_conf_file = "$args{'configdir'}/sanoid.defaults.conf";
+my $default_conf_file = "/usr/share/sanoid/sanoid.defaults.conf";
 
 # parse config file
 my %config = init($conf_file,$default_conf_file);
````

EDIT - verified Ubuntu  uses Debian source

gnordli:
Hi Jim. 
I don't think the Deb/Ubuntu package gets updated frequently enough.  I build from the master source and install the package.   I also put the package on hold.  

Geoff

jim-perkins:
thanks for the reply.
yes Deb/Ubuntu is slower to update.  In this case at the time of my install they happen to be at the same release level as the source (2.0.3)  I went with the deb to simplify my updates.  Sometimes I will use a master source but it isnt usually necessary for me.  I am hoping sanoid is a set and forget type of thing.

Regarding this issue and my attempt at a pull request:   Some deb users will blindly install and then eventually find a doc that states you have to copy the config over.  If they happen to come here for documentation they will be left scratching their head about the sanoid.defaults.conf.  

[...]

PR #577, opened 2020-08-03
Update README.md

jim-perkins:
Update README to reflect that Ubuntu stores 'default' file in an alternate location
https://github.com/jimsalterjrs/sanoid/issues/574


phreaker0:
@jim-perkins This is a bit misleading, if you build the package as stated in the INSTALL.md, the location of the configuration files is the default one on debian/ubuntu. I don't think it's a good idea to document changes from different distribution methods here.

PR #535, opened 2020-04-28
Fixed SSH control socket collision

sasoiliev:
Fixes #532

Issue #532, opened 2020-04-21
[syncoid] CRITICAL ERROR: snapshots couldn't be listed


sasoiliev:
Hi,

I am trying to set up a set of periodic push replications from host A to host B with syncoid.

(The reason for not using the `--recursive` option is that there is no direct mapping of the dataset hierarchy on the two hosts, i.e. I can sync some of the datasets with recursion, but others I still want to map manually.)

I have created systemd service and timer units for each dataset pair. The timer units are configured to trigger at the same time (every hour).

I am hitting an issue where at least one (but usually more) of the syncoid instances fail to list the snapshots on the remote host.

Here's the debug log of the latest run of one of the instances:

```
Apr 21 01:30:00 cube syncoid[7046]: DEBUG: SSHCMD: ssh    -i /root/.ssh/id_rsa
Apr 21 01:30:01 cube syncoid[7046]: ControlSocket /tmp/syncoid-root-root@offsite-1587425400 already exists, disabling multiplexing
Apr 21 01:30:02 cube syncoid[7046]: DEBUG: checking availability of lzop on source...
Apr 21 01:30:02 cube syncoid[7046]: DEBUG: checking availability of lzop on target...
Apr 21 01:30:02 cube syncoid[7046]: DEBUG: checking availability of lzop on local machine...
Apr 21 01:30:02 cube syncoid[7046]: DEBUG: checking availability of mbuffer on source...
Apr 21 01:30:02 cube syncoid[7046]: DEBUG: checking availability of mbuffer on target...
Apr 21 01:30:03 cube syncoid[7046]: DEBUG: checking availability of pv on local machine...
Apr 21 01:30:03 cube syncoid[7046]: DEBUG: checking availability of zfs resume feature on source...
Apr 21 01:30:03 cube syncoid[7046]: DEBUG: checking availability of zfs resume feature on target...
Apr 21 01:30:03 cube syncoid[7046]: DEBUG: syncing source persistent/content to target backup/content.
Apr 21 01:30:03 cube syncoid[7046]: DEBUG: getting current value of syncoid:sync on persistent/content...
Apr 21 01:30:03 cube syncoid[7046]:   zfs get -H syncoid:sync 'persistent/content'
[...]

PR #581, opened 2020-08-20
include PID in SSH Control path

human-capitalist:
The getssh() function generated a control path using the target host, ssh username, and time() function, which only has 1 second resolution.

This can lead to problems with multiple separate sync jobs running concurrently within 1 second of each other, e.g. when backing up multiple filesystems via systemd services. Both jobs get the same ControlPath and end up using the same multiplexed ssh socket, that is then closed by whichever job finishes first, causing unpredictable errors in the other job when it's stdout or stdin pipe is cut.

The change inserts the PID of the Perl process into the ControlPath to prevent separate processes interfering with each other when running in parallel with less than 1 second between them.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1433,7 +1433,7 @@ sub getssh {
 		 $remoteuser =~ s/\@.*$//;
 		if ($remoteuser eq 'root' || $args{'no-privilege-elevation'}) { $isroot = 1; } else { $isroot = 0; }
 		# now we need to establish a persistent master SSH connection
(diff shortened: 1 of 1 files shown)


human-capitalist:
So I should have looked at other pull requests first...
https://github.com/jimsalterjrs/sanoid/pull/535
seems like a more comprehensive fix than my one?
Presumably you can discard this PR if 535 is OK.

human-capitalist:
...as discussed in https://github.com/jimsalterjrs/sanoid/issues/532

PR #584, opened 2020-08-31
Tweak to INSTALL.md for Centos 8

Topslakr:
The PowerTools repo is required for Centos 8. To enable it you install dnf plugins first.

Additionally, you can add the '--now' option to a 'systemctl enable' command to enable _and_ start the item. One less command...

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -52,6 +52,9 @@ Install prerequisite software:
 ```bash
 # Install and enable epel if we don't already have it, and git too
 sudo yum install -y epel-release git
+# For Centos 8, install dnf plugins, and use them to enable the 'PowerTools' Repo
+sudo dnf -y install dnf-plugins-core
+sudo dnf config-manager --set-enabled PowerTools
 # Install the packages that Sanoid depends on:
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -136,8 +139,7 @@ sudo systemctl daemon-reload
 # Enable sanoid-prune.service to allow it to be triggered by sanoid.service
 sudo systemctl enable sanoid-prune.service
 # Enable and start the Sanoid timer
-sudo systemctl enable sanoid.timer
-sudo systemctl start sanoid.timer
+sudo systemctl enable --now sanoid.timer
```
Topslakr, line 142: Using 'enable --now' performs the same task as enable, and then start.

INSTALL.md
```diff
@@ -52,6 +52,9 @@ Install prerequisite software:
 ```bash
 # Install and enable epel if we don't already have it, and git too
 sudo yum install -y epel-release git
+# For Centos 8, install dnf plugins, and use them to enable the 'PowerTools' Repo
+sudo dnf -y install dnf-plugins-core
+sudo dnf config-manager --set-enabled PowerTools
```
Topslakr, line 57: These are needed for successful installation on Centos 8

PR #588, opened 2020-10-01
Syncoid only uses its own bookmarks

durkatlon:
Currently syncoid will use any bookmark it can find for a sync request. When setting a bookmark it does not leave behind an identifying mark. It never removes bookmarks. 

With these changes, bookmarks are handled much like the sync snapshots that syncoid sets (and cleans up when it's done with them). 

 - A bookmark that is set has _syncoid_${hostname} on the end of the name. 
 - Syncoid will only look for its own bookmarks when deciding which one to use.
 - When it's done using a bookmark, it is cleaned up.
 - Also handles bookmark to oldest snap then to newest snap.

I'm using this locally to create a 3rd backup of a 2nd-ary backup that is populated by pve-zsync (ProxMox) which uses zfs send/receive and would thereby wipe out the syncoid sync snaps from the tertiary backup box.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -757,6 +757,15 @@ sub syncdataset {
 						}
 					};
 				}
+				# This bookmark is no longer of use, delete it
+				my $deletebookmarkcmd = "$sourcesudocmd $zfscmd destroy $sourcefsescaped#$bookmarkescaped";
+				if ($sourcehost ne '') {
+					$deletebookmarkcmd = "$sshcmd $sourcehost " . escapeshellparam($deletebookmarkcmd);
+				}
+				if ($debug) { print "DEBUG: $deletebookmarkcmd\n"; }
+					system($deletebookmarkcmd) == 0 or do {
+					warn "CRITICAL ERROR: $deletebookmarkcmd failed: $?";
+				}
 			}
 
 			# do a normal replication if bookmarks aren't used or if previous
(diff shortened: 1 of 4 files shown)


phreaker0 (changes requested):
Hi, I implemented the bookmark stuff and it's designed to use any available bookmark to guarantee a working replication.

durkatlon:
> Hi, I implemented the bookmark stuff and it's designed to use any available bookmark to guarantee a working replication.

How about if I changed it like this:

- Bookmarks set by syncoid itself will still follow my new naming convention.
- For replication it will preferentially look for a bookmark set by syncoid from the current host. However it will use any available bookmark, if such a bookmark cannot be found.
- *If* a syncoid bookmark belonging to the host that the script is running on is used as the source of a replication step, it subsequently gets deleted. 
- If a non-syncoid bookmark is used, it does not get deleted.
- If a syncoid bookmark belonging to a different host is used, it does not get deleted.

This would still satisfy my main requirement of not leaving behind an ever-growing stack of bookmarks, and it should still satisfy your requirement that any available bookmark is used to perform the sync.

What do you think?

0xFelix:
I'm backing up irregularly to multiple host and your proposal would fit my use case perfectly, too. 

Currently backing up irregularly to more than two destinations prints a scary warning (CRITICAL ERROR) that a bookmark could not be created because a bookmark without suffix and a bookmark with a guid suffix already exist. The bookmarks identify the exact same source snapshot and so no harm is done and the replication succeeds in spite of that. Being able to identify bookmarks could solve this.

jimsalterjrs:
Bookmarks, unlike snapshots, are effectively free: there's nothing *in* a bookmark but a single TXG ID that tells replication "I don't have this snapshot, but my replication target *does* still have this snapshot, so if you just send all the blocks you have from TXGs newer than the one, the remote system will be able to base an incremental receive on that."

I might consider bookmark pruning in sanoid, maybe, but definitely not in syncoid, sorry. There's just no real gain, while there's significant potential for accidental breakage.

Issue #153, opened 2017-10-20
syncoid exclude snapshot types by name


graham00:
I've made some changes to add a new command, --exclude-snaps, so that I can avoid replicating certain types of snapshots. For example:

syncoid --exclude-snaps="hourly"

This skips snapshots in getsnaps() matching that term, and it then loops syncdataset with -i from snapshot to non-excluded snapshot until it gets a return value indicating it reached the last snapshot. Seems to be cutting the amount of replicated data for most of my VMs over WAN down to less than 50% of what it was.

I can make a pull request with the changes, but if you're not wanting to approach this issue from that direction, then I'll probably just add it in my [rsync fork](https://github.com/graham00/syncoid-rsync).

sithson:
Hi Graham and jimsalterjrs.

I would also love to see the --exclude functionality as I don't need to replicate hourly snapshots to my backup site.

As of now, hourly snaps always replicate, then sanoid removes them every minute on the backend site.
Very annoying, and causes unnecessary stress on my backup disks.

Thanks for a great tool jimsalterjrs, and I hope you consider Grahams request =)
Cheers!

phreaker0:
@graham00 did you create a branch/pr anywhere with your changes?

redmop:
Does Sanoid handle not trying to delete snapshots that are cloned? That would be a handy addition.

phreaker0:
@redmop zfs destroy will error out in this case by default:

```
zfs destroy test/a@test
cannot destroy 'test/a@test': snapshot has dependent clones
use '-R' to destroy the following datasets:
test/clone
```

redmop:
So I'll receive this error every time sanoid runs until excluded or it can actually delete it. Maybe a patch to include an option to ignore this error?

phreaker0:
@redmop A patch to silence would be possible but this would need to be configurable which should default to off. In the normal use case if you instruct sanoid to prune specific snapshots you want to know if it failed.

redmop:
Default to off, of course.
I could always just rename the snapshot when I clone it too.

phreaker0:
@graham00 ping

graham00:
> @graham00 did you create a branch/pr anywhere with your changes?

@phreaker0 I did - see the link to my fork in my opening comment. The usage for excluding hourly snapshots is documented. If you're only after that, then I suggest you also add "--no-rsync" as is described above that section, so that my fork otherwise behaves like the original syncoid does.

[...]

PR #593, opened 2020-10-14
Added support for exclude-snaps

RubyFeinstein:
Regarding https://github.com/jimsalterjrs/sanoid/issues/153

I've made some changes to add a new command, --exclude-snaps, so that I can avoid replicating certain types of snapshots. For example:

syncoid --exclude-snaps="hourly"

This skips snapshots in getsnaps() matching that term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,14 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 	The given regular expression will be matched against all datasets which would be synced by this run and excludes them. This argument can be specified multiple times.
 
++ --exclude=REGEX
+
+	The given regular expression will be matched against all datasets which would be synced by this run and excludes them. This argument can be specified multiple times.
+
++ --exclude=snaps=REGEX
+
+	The given regular expression will be matched against all snapshots which would be synced by this run and excludes them. This argument can be specified multiple times.
+
 + --no-resume
 
 	This argument tells syncoid to not use resumeable zfs send/receive streams.
--- a/syncoid
+++ b/syncoid
@@ -23,7 +23,7 @@ my $pvoptions = "-p -t -e -r -b";
 my %args = ('sshkey' => '', 'sshport' => '', 'sshcipher' => '', 'sshoption' => [], 'target-bwlimit' => '', 'source-bwlimit' => '');
 GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsnaps", "recursive|r", "sendoptions=s", "recvoptions=s",
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
-                   "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
+                   "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "exclude-snap=s@", "skip-parent", "identifier=s",
                    "no-clone-handling", "no-privilege-elevation", "force-delete", "no-clone-rollback", "no-rollback",
                    "create-bookmark", "pv-options=s" => \$pvoptions,
                    "mbuffer-size=s" => \$mbuffer_size) or pod2usage(2);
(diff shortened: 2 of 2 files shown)


redmop:
Missing documentation

shoote:
Updated README.md with the new syncoid flag, I can edit the wiki once it's merged.
Also wanted to note that I'm not 100% familiar of the code base, so hoping my naive implementation is good enough.

redmop:
Does it exclude intermediate snapshots, or only the end ones?

If I use syncoid on the following list of snapshots, what would I end up with?

rpool@exclude-1
rpool@keep-1
rpool@exclude-2
rpool@keep-2
rpool@exclude-3

shoote:
Assuming no snapshot exists on target, it will run:
```bash
sudo zfs send  'pool-test'@'keep-1' 
sudo zfs send  -I 'pool-test'@'keep-1' 'pool-test'@'keep-2'
```

On the target you will end up with:
```
NAME                       USED  AVAIL     REFER  MOUNTPOINT
pool-test/test@keep-1       14K      -       24K  -
pool-test/test@exclude-2    14K      -       24K  -
pool-test/test@keep-2        0B      -       24K  -
```

On a side note, I wanted to use `--no-stream` to avoid sending the intermediate snapshot, but it only sends the latest snapshot.
Might be cool to add another flag like --no-intermediary that will result in running:
```bash
sudo zfs send  'pool-test'@'keep-1' 
sudo zfs send  -i 'pool-test'@'keep-1' 'pool-test'@'keep-2'
```

redmop:
Sounds good. I'd put that in the docs.

mreymann:
Hi shoote,

Using your version of syncoid from https://github.com/singular-labs/sanoid/tree/exclude_snap we still have "hourly" snapshots in the replication target.
This is the command line we use:

`$ /opt/sanoid/syncoid_exclude_snap zones root@10.1.73.200:dpool01/smart03/zones --debug --exclude-snaps="hourly" --no-sync-snap --recursive > /opt/sanoid/logs/syncoid.log`

The log says that it excludes hourly snaps and it doesn't say it sends hourly snaps:

```
[16:13:12][root@smart03:/opt/sanoid]$ grep hourly /opt/sanoid/logs/syncoid.log | head -3
DEBUG: excluded snapshot autosnap_2020-10-15_14:00:02_hourly because of hourly
DEBUG: excluded snapshot autosnap_2020-10-15_15:00:02_hourly because of hourly
DEBUG: excluded snapshot autosnap_2020-10-15_16:00:01_hourly because of hourly
```

Still, the snaps end up in the replication target:

```
[14:57:10][root@repli-storage:~]$ zfs list -t snapshot | grep "dpool01/smart03/" | grep hourly
dpool01/smart03/zones@autosnap_2020-10-15_14:00:02_hourly                                                                             0B      -      648K  -
dpool01/smart03/zones@autosnap_2020-10-15_15:00:02_hourly                                                                             0B      -      648K  -
dpool01/smart03/zones@autosnap_2020-10-15_16:00:01_hourly                                                                             0B      -      648K  -
dpool01/smart03/zones@autosnap_2020-10-15_17:00:02_hourly
...
```

Are we using your version correctly?

redmop:
It only excludes snapshots from being the first and last. It does not skip ones in the middle. See my example above.
To do skip all snapshots matching the expression, it would need to run a zfs send -i for each snapshot not excluded.

shoote:
That's right, which means that it makes --exclude-snaps useful only when combining with --no-stream flag.
@redmop do you think we should change that to implement the loop you suggested?

redmop:
I would strongly suggest the loop. It is what I would expect when I use --exclude-snaps. Otherwise, there will be 1200 tickets made because it isn't working like people expect (excluded snapshots in the middle still being sent).

I think the loop is really going to slow things down, though.

I currently do it like this: I sync the whole thing, then remove the offending snapshots on the destination. But I have the space and bandwidth to do so.

shoote:
sure, I will try to find the time to implmenet the loop. will change the pull request to draft in the meantime.

redmop:
I'm not a dev of sanoid/syncoid, so I'd wait for one of them to respond.

@jimsalterjrs @phreaker0 (I feel like I am missing some, but this is who I can remember right now)

phreaker0:
@shoote hi, thanks for tackling this and yes we need the loop as this is the expected behaviour.

NiklasGollenstede:
In it's current state, this actually does pretty much exactly the same as my slightly earlier pull request #567.

The only difference is that `--accept-snap` uses a single RegEx as whitelist, and that the naming tries to avoid the implication that the snaps will be actually (always) in-/excluded from the backup: it just whether they are accept(able) as explicit send targets.

And (the description of) #567 also gives another motivation why one would use the flag (as is, i.e. with incremental send): push replication without granting the remote/source user delete access on the backup target.

graham00:
I can't do a pull request since I already have a quite divergent fork, but I've brought my very outdated fork's "--exclude-snaps" code over into a copy of the current version of syncoid. See the comment here:

[https://github.com/jimsalterjrs/sanoid/issues/153#issuecomment-809638516](https://github.com/jimsalterjrs/sanoid/issues/153#issuecomment-809638516)

> yes we need the loop as this is the expected behaviour.

The way I've got --exclude-snaps set up, it does this - it loops and uses successive zfs send -i between non-excluded snapshots only so that it skips any undesired incrementals.

phreaker0:
implemented with #699

PR #605, opened 2020-12-07
implemented removal of conflicting snapshots with force-delete option

phreaker0:
the following scenario will work with this PR:

```
zfs create pool/dataset
zfs snapshot pool/dataset@duplicate
syncoid pool/dataset pool/dataset2
zfs destroy pool/dataset@duplicate
zfs snapshot pool/dataset@duplicate
syncoid --force-delete pool/dataset pool/dataset2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,11 +336,11 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 
 + --force-delete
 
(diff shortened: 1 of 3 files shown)


phreaker0:
@jimsalterjrs can we get this merged? i used it in production since implementing it.

PR #606, opened 2020-12-09
fix duplicate key definitions by only using the first occurence and p…

### LINKED ISSUE #604: Configuration does not deal well with duplicate lines
If a boolean in a configuration is defined it does not behave as expected. For example if the line "autosnap = no" is in a template it behaves as expected, however if it is in the template twice then it enables autosnap.

This is a problem because people often copy/paste between configs and in a line by line review everything is correct as both lines set the same value.

phreaker0:
…rinting a warning

Fixes #604


phreaker0:
@jimsalterjrs can we get this merged? it's a really simple fix and is got quite some age now.

PR #626, opened 2021-03-14
Add option to ignore failed bookmark creation

0xFelix:
When replicating to multiple targets (>2) bookmark and bookmark with
suffix will already have been created. This option can be used to ignore
creation failure of further bookmarks.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -25,8 +25,8 @@ GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsn
                    "source-bwlimit=s", "target-bwlimit=s", "sshkey=s", "sshport=i", "sshcipher|c=s", "sshoption|o=s@",
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
                    "no-clone-handling", "no-privilege-elevation", "force-delete", "no-clone-rollback", "no-rollback",
-                   "create-bookmark", "pv-options=s" => \$pvoptions, "keep-sync-snap", "preserve-recordsize",
-                   "mbuffer-size=s" => \$mbuffer_size) or pod2usage(2);
+                   "create-bookmark", "ignore-failed-create-bookmark", "pv-options=s" => \$pvoptions, "keep-sync-snap",
+                   "preserve-recordsize", "mbuffer-size=s" => \$mbuffer_size) or pod2usage(2);
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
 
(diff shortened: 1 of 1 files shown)


0xFelix:
@jimsalterjrs @phreaker0 Ping

phreaker0:
I think a better way to handle this would be to check the bookmark on failure if it already exists and points to the correct snapshot and if not still fail. So there wouldn't be the need for an extra argument and one can be sure that the bookmark is correct.

0xFelix:
@phreaker0 Makes sense! Updated the PR to implement that.

syncoid
```diff
@@ -837,15 +837,16 @@ sub syncdataset {
 	}
 	if (defined $args{'no-sync-snap'}) {
 		if (defined $args{'create-bookmark'}) {
-			my $ret = createbookmark($sourcehost, $sourcefs, $newsyncsnap, $newsyncsnap);
+			my %existingbookmarks = getbookmarks($sourcehost,$sourcefs,$sourceisroot);
+			my $guid = $snaps{'source'}{$newsyncsnap}{'guid'};
+			my $ret = createbookmark($sourcehost, $sourcefs, $newsyncsnap, $newsyncsnap, $guid, %existingbookmarks);
 			$ret == 0 or do {
```
phreaker0, line 843: this code should stay as it where before, because doing it this way will degrade performance because each time all the bookmarks needs to be listed.
It's better to handle that in the fallback case (if the bookmark creation fails), there the bookmark listing and the compare if this bookmark already exists should be made.

syncoid
```diff
@@ -1066,7 +1067,12 @@ sub syncbookmark {
 } # end syncbookmark
 
 sub createbookmark {
-	my ($sourcehost, $sourcefs, $snapname, $bookmark) = @_;
+	my ($sourcehost, $sourcefs, $snapname, $bookmark, $guid, %existingbookmarks) = @_;
+
+	if (defined $existingbookmarks{$guid} && $existingbookmarks{$guid}{'name'} eq $bookmark) {
+		writelog('INFO', "bookmark already exists, skipping creation");
+		return 0;
+	}
```
phreaker0, line 1075: this should be move to the fallback case if the bookmark creation failed

tests/syncoid/12_bookmark_check_guid_on_existing/run.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
phreaker0, line 1: because of #904 this needs to be renamed with a 0 prefix.

phreaker0 (changes requested):

0xFelix:
Same here, here's yet another revision of this PR. I've addressed all your comments.

@phreaker0 I hope we can get this merged now?

0xFelix:
Ping @phreaker0 @jimsalterjrs

phreaker0:
@0xFelix i tested it and looks good! thx for your work

PR #632, opened 2021-03-31
Added --exclude-snaps feature

graham00:
This adds the --exclude-snaps feature that I have used in my own fork for years back to the current copy of syncoid.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -4,7 +4,7 @@
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
 # project's Git repository at https://github.com/jimsalterjrs/sanoid/blob/master/LICENSE.
 
-$::VERSION = '2.0.3';
+$::VERSION = '2.0.4';
 
 use strict;
 use warnings;
@@ -145,6 +145,7 @@ my %avail = checkcommands();
 
 my %snaps;
 my $exitcode = 0;
+my $errorcount = 0;
 
 ## break here to call replication individually so that we ##
 ## can loop across children separately, for recursive     ##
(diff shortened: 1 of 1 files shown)


zwets:
@graham00 thanks for submitting the PR, this would indeed be a great feature. However it looks like it needs some more testing. I tried this:

    zfs create rpool/source
    zfs snapshot rpool/source@keepme
    zfs snapshot rpool/source@dontkeepme
    syncoid --exclude-snaps=dont rpool/source rpool/dest

This correctly results in `rpool/dest` having the `@keepme` snapshot and not the `@dontkeepme`. However it also doesn't get the `@syncoid` snap I would expect it to have, whereas `rpool/source` ends up with two `@syncoid` snaps. Every subsequent sync results in another `@syncoid` snap at the source.

I also had an issue when I ran with `--exclude-snaps=@syncoid.*` (a perfect use case for this feature). This seemed to make the sync go into a tight loop, creating a `@syncoid` snap on `rpool/source` every second, but nothing on the target.

I suspected this was related to these lines in the source code (they occur multiple times):

    foreach my $term (@excludeterms) {
        if ($line =~ /$fs\@.*$term/) {
            $skipsnap = 1;
        }
    }

    # For now, also assume we don't want to sync any @syncoid* produced by sync-snap if we are excluding autosnaps ... could always break this out into another option like --exclude-sync-snaps later if needed.
    if ($line =~ /$fs\@syncoid.*/) {
        next;
    }
    if ($skipsnap) {
        $skipsnap = 0;
        next;
    }

Note that a match means that the loop will be reexecuted with `$snapshot` still being 1 (as the third block is never reached), which I doubt is the intention.

Removing the `# For now ...` section altogether (to give the user the choice whether to sync the `@syncoid`) didn't resolve the issue either.

It looks to me like your code is almost there. I wish I had a bit more time to dig into this.

phreaker0:
implemented with #699

PR #636, opened 2021-04-08
Implementing support for ssh_config(5) files

endreszabo:
I implemented the support to use the `-F` ssh client argument. This allows using a custom `ssh_config(5)` file, which my setups rely on heavily. Tested working :)


phreaker0:
@jimsalterjrs Looks good, there is also another older one:

Closes #495

PR #667, opened 2021-08-16
Sort snapshots by createtxg

shodanshok:
Sort by snapshot creation can be subject to collisions - by using single-second resolution, two consecutive snapshots can very well have the *same* creation time. `syncoid` already has some protection from this issue, as it uses a 3-digit code/suffix to correctly enumerate snapshots with same creation time. 

Still, the issue can be completely avoided sorting snapshots by `createtxg`. From [zfsprops man page](https://openzfs.github.io/openzfs-docs/man/7/zfsprops.7.html):

> **createtxg**
> The transaction group (txg) in which the dataset was created. Bookmarks have the same createtxg as the snapshot they are initially tied to. *This property is suitable for ordering a list of snapshots, e.g. for incremental send and receive.*

This patch enable sorting snapshots by `createtxg`. As bookmarks have the same txg number than their parent snapshots, I did not change bookmark sorting/matching. 

FULL DISCLOSURE: I was using an ancient version of `syncoid` (without suffix for same-time snapshots) on a running pair of systems which were recently updated to ZFS 2.0.5. Due to creation time collision, `syncoid` sometime tried to transfer an already-transferred snapshot to the target machine. Normally this should pose no issue, because ZFS simply ignores the duplicate snapshot and goes ahead. For this specific pools, however, something strange happens: transferring an already sent snapshot [kill the whole transfer](https://github.com/openzfs/zfs/issues/12476) with a `cannot receive: invalid record type` error, effectively stopping the replication. With the current `syncoid` script (and its added code/suffix for same-time snapshots) the issue *should* be mitigated, but I figured that sorting by creation txg can completely avoid it.

***This patch has received only minimal testing - it seems to works for me, and that's all***

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -608,7 +608,7 @@ sub syncdataset {
 			my %bookmarks = getbookmarks($sourcehost,$sourcefs,$sourceisroot);
 
 			# check for matching guid of source bookmark and target snapshot (oldest first)
-			foreach my $snap ( sort { $snaps{'target'}{$b}{'creation'}<=>$snaps{'target'}{$a}{'creation'} } keys %{ $snaps{'target'} }) {
+			foreach my $snap ( sort { $snaps{'target'}{$b}{'createtxg'}<=>$snaps{'target'}{$a}{'createtxg'} } keys %{ $snaps{'target'} }) {
 				my $guid = $snaps{'target'}{$snap}{'guid'};
 
 				if (defined $bookmarks{$guid}) {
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
Can we get some volunteers to test, please?

Also, does anybody know how long `createtxg` has been an available property? I don't *think* it was there several versions ago... but I might be wrong. If we implement something that will break users on older versions, we need to be aware of the scope.

shodanshok:
@jimsalterjrs `createtxg` should have been introduced alongside bookmarks with ZFS 0.7.x, so the only pool impacted would be the ones with 0.6.x (or zfs-fuse). If supporting such old pools is important/valuable, we can add some auto-detection code to select the correct order field.

PR #698, opened 2021-11-28
feat: Samba friendly snapshot name

kimata:
Renaming a snapshot to make it easier to handle with Samba's vfs_shadow_copy2.

After applying the above changes, the following settings in smb.conf should work.

```
     shadow:format       = _UTC_%Y-%m-%d_%H:%M:%S                                                                                                                                                          
     shadow:snapprefix   = ^autosnap_\(yearly\)\{0,1\}\(monthly\)\{0,1\}\(weekly\)\{0,1\}\(daily\)\{0,1\}\(hourly\)\{0,1\}\(frequently\)\{0,1\}                                                            
     shadow:delimiter    = _UTC_                                                                                                                                                                           
     shadow:localtime    = no
```

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -546,7 +546,7 @@ sub take_snapshots {
 			my @snapshots;
 
 			foreach my $type (@types) {
-				my $snapname = "autosnap_$datestamp{'sortable'}_$type";
+				my $snapname = "autosnap_${type}_UTC_$datestamp{'sortable'}";
 				push(@snapshots, $snapname);
 			}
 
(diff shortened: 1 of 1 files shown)


stumpylog:
This would be great to have included in sanoid.  I use ZFS and share out some datasets using Samba.

I wonder if there needs to be some sort of rename of existing snapshots? To facilitate the pruning of old named snapshots?

sporkus:
This and removing the colons would help windows users. vfs_shadow_copy2 does not allow regex/wildcard after the time string.

natankeddem:
What would it take to get this feature actually implemented? I am going to just spitball some ideas:

- Make it enabled with an argument not a blanket change for everyone to preserve backward compatibility
- remove colons from date since that seems to cause some people issues as well (although I haven't experienced this tbh)
- ???

phreaker0:
@kimata we can't change the default behaviour. But there is now a PR where it can be configured freely: #989

PR #710, opened 2022-01-16
Twice prepended $shcmd in $rhost for getsnapsfallback

NOVACyclist:
getsnapsfallback is only called from within getsnaps if an error is detected in the command. The block:

if ($rhost ne '') {
$rhost = "$sshcmd $rhost";
# double escaping needed
$fsescaped = escapeshellparam($fsescaped);
}

Is used in both functions getsnapsfallback and getsnaps. The issue is that syncoid executes this block once in getsnaps, then when passed to getsnapsfallback it is always executed again on $rhost (as it is impossible for $rhost to be null at entry) causing the $sshcmd to be prepended twice to the command string, permanently breaking this function for all hosts that failed getsnaps.

As this block will have already been executed by getsnaps, and getsnapsfallback isn't called anywhere else in the program I am proposing removing the duplicate block from getsnapsfallback instead of writing tests here. Any additional tests/checks should be written into the main block to be inherited into this child call. Tested and validated remote send/recv execution on Solaris (5.10 10/09 -> 5.11 11.4).

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1596,12 +1596,6 @@ sub getsnapsfallback() {
 	my $fsescaped = escapeshellparam($fs);
 	if ($isroot) { $mysudocmd = ''; } else { $mysudocmd = $sudocmd; }
 
-	if ($rhost ne '') {
-		$rhost = "$sshcmd $rhost";
-		# double escaping needed
-		$fsescaped = escapeshellparam($fsescaped);
-	}
-
 	my $getsnapcmd = "$rhost $mysudocmd $zfscmd get -Hpd 1 type,guid,creation $fsescaped |";
 	warn "snapshot listing failed, trying fallback command";
 	if ($debug) { print "DEBUG: FALLBACK, getting list of snapshots on $fs using $getsnapcmd...\n"; }


mr-vinn:
The problem with removing that entire block in `getsnapsfallback()` is that you won't double-escape the filesystem name, since `getsnaps()` doesn't pass in the already-escaped value.  You could pass the escaped value so that you only have to do it in `getsnaps()`, in which case you should also update `getsnapsfallback()` so that you don't end up escaping the filesystem a third time on line 1596.

NOVACyclist:
> The problem with removing that entire block in `getsnapsfallback()` is that you won't double-escape the filesystem name, since `getsnaps()` doesn't pass in the already-escaped value. 

Yes it does, the altered values are already being passed which is why I'm suggesting the removal of the duplicated block. Should there be another entry condition to test for it should be done here and not later. Should there be a need to alter these variables differently for the `getsnapsfallback()`, than in `getsnaps()` the variables should be modified into local working copies per-sub which is not what is being done. 

Presently in `getsnaps()`, the block:
`if ($rhost ne '') {
  			$rhost = "$sshcmd $rhost";
  			# double escaping needed
  			$fsescaped = escapeshellparam($fsescaped);
  		}`
 
is called directly after the variables are initiated. Because the entry condition of the block is only`$rhost ne ''` the entry condition will be met every additional time this block is encountered. Because this block directly modifies both `$rhost` and `$fsescaped` it will prepend `$sshcmd` to $rhost every time it is called as well as double escaping all single quotes which would also likely cause issue.

mr-vinn:
> the altered values are already being passed

That's true for `$rhost`, but not `$fsescaped`.  In `getsnaps()`, the escaped filesystem name is stored in `$fsescaped`.  But `$fs` is passed to `getsnapsfallback()`, and that value is not escaped.

NOVACyclist:
> > the altered values are already being passed
> 
> That's true for `$rhost`, but not `$fsescaped`. In `getsnaps()`, the escaped filesystem name is stored in `$fsescaped`. But `$fs` is passed to `getsnapsfallback()`, and that value is not escaped.

Thanks for that catch. Updated to pass values and not duplicate efforts in the fallback function.

phreaker0:
fixed with #767, thx but unfortunately i overlooked this PR

PR #739, opened 2022-04-13
Add sudo to last Debian install step

jimmyface:
The last install step for Debian involves installing the generated package via apt. However, this requires sudo (at least by default.) Add sudo to the last step to make each step work properly.


phreaker0:
already in with #690, thx

PR #746, opened 2022-05-19
Removed unneeded 2>/dev/null from mbufferoptions

mavhc:


phreaker0:
read a bit of documentation and the redirect shouldn't be needed as mbuffer shouldn't print anything on stderr in the normal case with the quiet flag (without stats data is printed to stderr).

Issue #11, opened 2015-08-11
Dry Run Options


gdevenyi:
I want to see that my configuration is correct, or see what syncoid will do before I do it. A dry run option would be nice.

mailinglists35:
i'd like to see this, too! :)

mailinglists35:
hi,
just a reminder
thanks

mailinglists35:
hey guys, is this thing difficult to implement?

Zixim:
this would be great to have.

PR #748, opened 2022-05-30
implement --dryrun argument

anarcat:
This is a little messy: that includes a *lot* of whitespace change,
but I couldn't really figure out a better way to do it, in general.

(A better way would have been something like:

    if ($args{'dryrun'}) return 0;

... but the functions are generally too long for that kind of hack to
work at all. Alternatively, I figured it might be good to have a
generic `system()` wrapper, but I haven't dug deep enough in the code
to see if it's always called the same way, or if that would make
sense.)

This doesn't block *all* `system()` calls either: only those who
actually make changes. SSH setup commands and commands that just probe
the snapshot lists are let through, so this is not completely a
`--noop` flag (hence the `--dryrun` instead), because it
actually *does* some things still.

I have only grepped for `system()` and haven't audited the entire
source code to see if other side effects (e.g. creating files) also
exist.

Disclaimer: this patch was briefly tested on my home system, and my
level of familiarity with sanoid is "beginner".

Closes: #11


phreaker0 (changes requested):
As you guessed, the white space issue needs to be resolved.
Did you test you change extensively because I guess there will be error's in edge cases for example with clone handling and such stuff.

anarcat:
i didn't test this very much, just a few trial runs. this is mostly a proof-of-concept I guess, looking for feedback whether this is a desirable feature at all.

jimsalterjrs:
Closed for inactivity.

PR #760, opened 2022-08-12
add --syncstats option

rwarren:
The `--syncstats` option is to support general per-dataset instrumentation of the syncoid process, and allows you to specify a file path to dump per-dataset stats to.  The specified file will contain json with stats for each dataset that is considered.

The original use case driving this change is to parse the output file after the sync is complete, and send the data to your metric tracking system/TSDB of choice.

For this commit, the only stat currently added is the sync duration (`sync_duration_s`).  However, the code/output has been set up to make it easy to add more per-dataset stats in the future.  For example, `delta_bytes` (the size of the incremental delta in bytes) would be another nice stat to add later.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -10,7 +10,9 @@ use strict;
 use warnings;
 use Data::Dumper;
 use Getopt::Long qw(:config auto_version auto_help);
+use JSON::PP;
 use Pod::Usage;
+use Time::HiRes;
 use Time::Local;
 use Sys::Hostname;
 use Capture::Tiny ':all';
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
@phreaker0 does this look worth revisiting to you? I think we kinda ignored it when it showed up because we had a lot of rather larger feature upgrades to worry about. It needs rebasing now, but I don't know if @rwarren is still around, and I'd like a second opinion on whether this is something worth resurrecting now. 💚

rwarren:
Still here, just not tracking this much anymore.  Stats would most definitely still be useful here, if not added yet.

phreaker0:
@jimsalterjrs i think this is a niche feature but it's not intrusive and on the simple side so why not.

@rwarren can you rebase you changes? There are some special cases like "sendoptions=R" where recursion is handled by zfs which this doesn't consider but for simplicity I would also keep it that way.

PR #762, opened 2022-08-22
Add preserve-compression option

tub5ta:
Datasets can have different compression options depending on the type of data stored, similar to the reason for having unique record sizes. This allow the preservation of the compression type set on a data set when replicating recursively.


phreaker0:
see #763

PR #763, opened 2022-08-22
Add --preserve-compression option

tub5ta:
This option allows the compression type to be preserved from the source host to the destination. Useful where there are different compression types used in various datasets that are being replicated as part of a recursive tree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,10 @@ As of 1.4.18, syncoid also automatically supports and enables resume of interrup
 + --preserve-recordsize
 
 	This argument tells syncoid to set the recordsize on the target before writing any data to it matching the one set on the replication src. This only applies to initial sends.
+	
++ --preserve-compression
+
+	This argument tells syncoid to set the compression type on the target before writing any data to it matching the one set on the replication src. This only applies to initial sends.
 
 + --no-clone-rollback
 


dodexahedron:
I'd suggest a more generic options flag, which gets passed to the `-o` parameter of `zfs send`, so that any number of options desired can be set. Since syncoid is often used for snapshot replication in a backup type scenario, it seems logical that the destination may often be at a _higher_ compression setting than the source.

It's difficult to find a definitive answer in documentation for the behavior, but my limited testing has shown that the compression of the source is maintained on the receiver, regardless of the compression setting on the receiver. The way I tested was with source and destination with zstd-4 and zstd-14 respectively. When a send/receive (without -c flag) is performed, the compression ratio and physical data size on the destination matches the source. When I use `zfs receive -o compression=zstd-14`, the ratio is higher and the data smaller. This tracks with how sends are supposed to send the properties of the source snapshot, needing a -o to override them at the destination.

Allowing a generic send/receive options flag would allow for any mix to be applied, along with whatever other options one may want to set (different hash algorithms, etc).

Thoughts?

phreaker0:
@dodexahedron this is already implemented, see https://github.com/jimsalterjrs/sanoid/blob/ef601eb96b016786af63ea5d7e30c8ad1b5de689/syncoid#L1961

phreaker0:
@tub5ta seems you missed the actual code

tub5ta:
Reviewing that section, it appears I have and I can revert my changes in my fork. I was looking for something simple like other 'preserve' flags when I was reviewing the code and missed this bit you have highlighted.

You can close this. If I hit a bug using the existing feature, I'll open a new PR. Thanks.

phreaker0:
k, thx

Issue #756, opened 2022-07-21
Introduce flag --no-forced-receive


msladek:
Recent [changes](https://github.com/jimsalterjrs/sanoid/commit/ef2b4f887a7cd1e139a500f1cac44e476b9ea198) to the master branch removing rollbacks also removed the `--no-rollback` flag, which additionally disabled [forced receive -F](https://openzfs.github.io/openzfs-docs/man/8/zfs-recv.8.html#F).

This flag was useful to prevent any possible destruction of snapshots on the target by syncoid. To avoid feature regression we could introduce a `--no-forced-receive` flag which removes the `-F` flag from the receive call.

PR #766, opened 2022-09-12
bring back no-rollback option

phreaker0:
this brings back the --no-rollback option without reverting to the old way of explicit rolling back the target filesystem.
The option is very useful to others (https://www.reddit.com/r/zfs/comments/xcffnn/syncoid_norollback_option_gone/) and me.


msladek:
@phreaker0 I've already adressed this issue with #756 a few months ago.
It instead names the flag `--no-forced-receive`, which now more accurately describes its effect. Please consider it before any merge.

phreaker0:
@msladek but this will break existing scheduled syncoid setups because the option is different. I would prefer not to break backward compatibility here as it's not needed imho.

msladek:
@phreaker0 that makes sense, yet for all those using the `master` branch backwards compatibility was already broken.

I will close my PR/Issue when this one is merged.

Majiir:
`--no-clone-rollback` was also removed. Is it safely replaced with the new `--no-rollback`?

Issue #815, opened 2023-04-13
Syncoid: Data loss because `getoldestsnapshot()` might not choose the first snapshot


Deltik:
## Bug Description
Due to an incorrect clock at the time, I have some snapshots where the `creation` property is in the past compared to earlier snapshots.

When `syncoid` performs an initial full transfer of a dataset, it could select a snapshot that isn't the first snapshot of the dataset because of [the usage of `getoldestsnapshot()`](https://github.com/jimsalterjrs/sanoid/blob/86fffca306ecd0e5bdb7eb413f4970024f96b53f/syncoid#L425).  This leads to data loss, as some snapshots may be left behind.

## How to Reproduce

See the bold parts of the output below:

<pre><code>root@box52:&#x7E;# truncate -s64M /tmp/jimsalterjrs_sanoid_815.img
root@box52:&#x7E;# zpool create jimsalterjrs_sanoid_815 /tmp/jimsalterjrs_sanoid_815.img
root@box52:&#x7E;# zfs create jimsalterjrs_sanoid_815/before
root@box52:&#x7E;# zfs snap jimsalterjrs_sanoid_815/before@this-snapshot-does-not-make-it-into-the-after-dataset
root@box52:&#x7E;# date --set="2006-08-14T02:34:56-06:00" ; zfs snap jimsalterjrs_sanoid_815/before@oldest-snapshot ; date
Mon Aug 14 03:34:56 AM CDT 2006
Mon Aug 14 03:34:56 AM CDT 2006
root@box52:&#x7E;# systemctl restart systemd-timedated.service
root@box52:&#x7E;# date
Thu Apr 13 11:38:36 AM CDT 2023
root@box52:&#x7E;# zfs snap jimsalterjrs_sanoid_815/before@another-snapshot-does-not-matter
root@box52:&#x7E;# syncoid --sendoptions="Lec" jimsalterjrs_sanoid_815/{before,after}
INFO: Sending oldest full snapshot jimsalterjrs_sanoid_815/before@oldest-snapshot (&#x7E; 12 KB) to new target filesystem:
42.8KiB 0:00:00 [6.57MiB/s] [=============================================================================================] 339%            
INFO: Updating new target filesystem with incremental jimsalterjrs_sanoid_815/before@oldest-snapshot ... syncoid_box52.deltik.net_2023-04-13:11:38:51-GMT-05:00 (&#x7E; 4 KB):
[...]

PR #818, opened 2023-04-25
syncoid: Initial sync: `getoldestsnapshot()` from the `createtxg` prop, not `creation`

Deltik:
It is possible for `creation` of a subsequent snapshot to be in the past compared to the current snapshot due to system clock discrepancies, which leads to earlier snapshots not being replicated in the initial syncoid sync.

Sorting snapshots by the `createtxg` value is reliable and documented in `man 8 zfsprops` as the proper way to order snapshots.

There are no other usages of `getoldestsnapshot()` besides this initial sync, so there is practically no regression potential of this change.

Fixes: https://github.com/jimsalterjrs/sanoid/issues/815


phreaker0:
@Deltik pls don't submit untested code changes. The object doesn't contain "createtxg", therefore this can't work and breaks the initial sync.

phreaker0:
Just saw that there is already a patch out there for using createtxg #667
Seems like not all versions of zfs have the createtxg property.

Deltik:
Ah, I'll need to take that into consideration and fall back to the original sorting by `creation`, then.  Thanks for the heads up!

Deltik:
Okay, @phreaker0, I've reworked the pull request with backwards compatibility in mind and added a test for the bugfix.

I'm not sure how to test OpenZFS 0.6, though, since the current test harness doesn't have the ability.

This pull request fully supersedes https://github.com/jimsalterjrs/sanoid/pull/667.

Deltik:
Rebased to be compatible with https://github.com/jimsalterjrs/sanoid/pull/809 (_support bookmarks which are taken in the same second_)

Deltik:
@jimsalterjrs: I've rebased this pull request to resolve conflicts with https://github.com/jimsalterjrs/sanoid/pull/699 and https://github.com/jimsalterjrs/sanoid/pull/859.  Could I get your review of this one so that I don't have to keep rebasing?

jimsalterjrs:
Sorry about the constant rebasing, @Deltik . I'm going to go ahead and merge; I assume you've been using your branch in prod, and you're obviously invested enough to come back if somebody discovers a serious issue with it in master. :)

phreaker0:
i just noticed a huge performance regression with this PR. Because now all properties are fetched instead of only the needed ones ( all vs guid,creation), I get why because if specifying createtxg it will fail for versions where it isn't supported but I guess we need to check for createtxg support once.

From one of my datasets:

$ time zfs get -Hpd 1 -t snapshot all '...' > /dev/null

real    0m2.642s
user    0m0.053s
sys     0m2.584s

$ time zfs get -Hpd 1 -t snapshot guid,creation '...' > /dev/null

real    0m0.804s
user    0m0.029s
sys     0m0.759s

Deltik:
I set up an old FreeBSD 8.3, which doesn't have <code>zfs get <strong>-t</strong></code>, and found two more regressions:

1. An oversight during [the merging of subroutine `getsnapsfallback` into `getsnaps`](https://github.com/jimsalterjrs/sanoid/pull/818/commits/e301b5b153a33d37f0ab062a90531226a718e1a4) caused the recursive call to prepend `$sshcmd` to `$rhost` again, leading to this error:

   ```shell
   # syncoid --debug root@192.168.122.26:rpool/before syncoid-test-11/after
   … [TRUNCATED] …
   WARNING: snapshot listing failed, trying fallback command
   DEBUG: getting list of snapshots on rpool/before using ssh      ssh      -S   /tmp/syncoid-root19216812226-1718239169-761069-2198 root@192.168.122.26  zfs get -Hpd 1   all ''"'"'rpool/before'"'"'' |...
   root@192.168.122.26: Command not found.
   CRITICAL ERROR: snapshots couldn't be listed for rpool/before (exit code 256) at /usr/sbin/syncoid line 1900.
   ```

2. If that error were fixed, the fallback command would still fail because non-snapshot datasets would be included in the output, leading to this parsing error:

   ```shell
   # syncoid --debug root@192.168.122.26:rpool/before syncoid-test-11/after
   … [TRUNCATED] …
   WARNING: snapshot listing failed, trying fallback command
   DEBUG: getting list of snapshots on rpool/before using ssh      -S   /tmp/syncoid-root19216812226-1718239740-763496-6973 root@192.168.122.26  zfs get -Hpd 1 all ''"'"'rpool/before'"'"'' |...
   CRITICAL ERROR: Unexpected dataset format in rpool/before       type    filesystem      - at /usr/sbin/syncoid line 1914.
   ```

I'm working on a fix for all three regressions.

Deltik:
Fixes are ready for review in https://github.com/jimsalterjrs/sanoid/pull/931!

PR #837, opened 2023-07-25
trim config values

### LINKED ISSUE #834: v2.2.0 autosnap=no appears broken in some cases
I moved to the recent v2.2.0 release and it seems that in this config `autosnap = no` is broken.

```
[nas]
        use_template = nas
        recursive = yes
        process_children_only = yes

[nas/rpool_backup]
        use_template = zfsreceive
        recursive = yes

[template_nas]
        hourly = 48
        daily = 30
        monthly = 12
        yearly = 0
        autosnap = yes
        autoprune = yes

[template_zfsreceive]
        autoprune = yes
        hourly = 48
        daily = 90
        monthly = 12
        yearly = 0

        ### don't take new snapshots - snapshots on backup
        ### datasets are replicated in from source, not
        ### generated locally
        autosnap = no

        ### monitor hourlies and dailies, but don't warn or
        ### crit until they're over 48h old, since replication
        ### is typically daily only
        hourly_warn = 2880
        hourly_crit = 3600
        daily_warn = 48
        daily_crit = 60
```

I expect an auto snap to not be taken for `nas/rpool_backup` and any children but one does appear every hour. This is then a problem when `syncoid` runs which sends/receives an `rpool` to `nas/rpool_backup` because now they both have identically named snapshots but with a different GUID which results in an error:
`cannot restore to nas/rpool_backup/misc@autosnap_2023-07-23_21:59:03_hourly: destination already exists`

sanoid + syncoid v2.1.0 work with the config above.

phreaker0:
leading and/or trailing spaces in the config values will lead to weird issues which are hard to track down, so trim them to get rid of those cases.

Fixes #834

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -905,6 +905,8 @@ sub init {
 				warn "duplicate key '$key' in section '$section', using the value from the first occurence and ignoring the others.\n";
 				$ini{$section}{$key} = $value->[0];
 			}
+			# trim
+			$ini{$section}{$key} =~ s/^\s+|\s+$//g;
 		}
 
 		if ($section =~ /^template_/) { next; } # don't process templates directly


timkgh:
Thank you so much. I've been running with a broken config for years because of this 😢

jimsalterjrs:
I can't believe this was necessary, tbh. We probably ought to file a bug upstream with the config::inifiles CPAN module.

timkgh:
Could you please do a 2.2.1 release? This is a pretty important fix. Thanks.

PR #845, opened 2023-08-09
Improve documentation

dlangille:
`recursive = zfs`

should be:

`recursive = yes`

### DIFF
--- a/sanoid.conf
+++ b/sanoid.conf
@@ -31,7 +31,7 @@
 # you can also handle datasets recursively in an atomic way without the possibility to override settings for child datasets.
 [zpoolname/parent2]
 	use_template = production
-	recursive = zfs
+	recursive = yes
 
 
 


jimsalterjrs:
Personally, I agree that recursive=yes is a better default. But this isn't a typo, it's two different methods of recursion. Recursive=zfs is more efficient, recursive=yes is more predictable (recursive=zfs won't take a monthly of a newly created child dataset until the parent also needs one; recursive=yes takes one right away).

I don't have much dog in the fight over which to default to, but a change needs discussion and consensus.

dlangille:
Sorry, I thought it was a typo.  I was sure it was because I didn't find a reference to `zfs` recursion.

dlangille:
Is this missing from the docs? I see `atomic` mentioned at https://github.com/jimsalterjrs/sanoid/blob/master/sanoid.conf#L31 but that wasn't enough to tell me `zfs` is different from `yes`

phreaker0:
no typo :-)
documentation pr's welcome :-)

dlangille:
Here I was, all ready to document if I could get questions answered....

phreaker0:
@dlangille sry. I thought we do the documentation update in a new PR but we can recycle this one too.

> Is this missing from the docs? I see `atomic` mentioned at https://github.com/jimsalterjrs/sanoid/blob/master/sanoid.conf#L31 but that wasn't enough to tell me `zfs` is different from `yes`

yes, that's all of the current documentation for recursive = zfs. Basically this runs zfs snapshot with the '-r' flag so zfs will recursively take a snapshot of the whole dataset tree which is consistent.

dlangille:
I added a new commit, and it involves some assumptions about `recursive = yes`, which I could not substantiate when reading the code. i.e. I could not see where that option does not use `-r`.

PR #855, opened 2023-10-15
debian: add openzfs-zfsutils as an alternative to zfsutils-linux

EchterAgo:
The package produced by ZFS 2.2.0 `make native-deb-utils` is called `openzfs-zfsutils`.


phreaker0:
thx

Issue #851, opened 2023-09-27
Feature request: New options to run the "sanoid --monitor-*" commands as non-root without updating the snapshot cache file


Pajkastare:
I use `sanoid` and `syncoid` for all my backup needs (and a bit of `rsync` from non-ZFS filesystems to ZFS), and I use `nagios` to monitor my local and remote ZFS status. All of this works great.

But: I'd like the ability to run `sanoid` as non-root for the `--monitor-snapshots`, `--monitor-capacity` and `--monitor-health` cases, **without** updating the snapshot cache file (`/var/cache/sanoid/snapshots.txt`, which by default times out after 15 minutes) even if it were to be too old. Suggested new option names are `--monitor-snapshots-no-update`, `--monitor-capacity-no-update` and `--monitor-health-no-update` for the new behaviour.

Well, to be precise, these commands _can_ be run as non-root today, but in case the snapshot cache file is too old, the commands will fail because `sanoid` automatically attempts to refresh the cache. (Perhaps it makes the most sense to return an "UNKNOWN" status code if it cannot be concluded if the snapshots are OK/WARN/CRIT, but the capacity and health commands should not need a cache file update.)

Why do I think this is a good idea? Not only because of the "too-old-snapshot-file issue" I described above, but also because the updating of the cache file causes unnecessary peaks in disk utilization, and unnecessary wear on mechanical HDDs. 
For example: A zpool of large HDD[s] only used for backup purposes, with snapshots sent a couple of times per day, doesn't need all snapshots indexed all the time.

What is the current behaviour? Just having Nagios look after the zpools by using the monitor commands above will result in "flapping services" (as it's called in Nagios), depending on how long after a systemd-timer-invoked `sanoid` execution as root occurs.
(My workaround is to add this to `/etc/crontab`: `*/14 * * * * root /usr/sbin/sanoid --force-update >/dev/null 2>&1`, which never gives me flapping Nagios services, but it also definitely causes unnecessary disk usage.) I guess I could, but I prefer not to, allow Nagios/NRPE to run `sanoid` as root. Or I could write a wrapper script that doesn't run the command if the snapshot file is too old. But I don't think I'm the only one who has this usecase.

I don't speak perl at all, so I don't think I should attempt to create a pull request for it.

jimsalterjrs:
[...]

PR #856, opened 2023-10-23
Fixes jimsalterjrs/sanoid#851

Pajkastare:
Based on text in the issue https://github.com/jimsalterjrs/sanoid/issues/851, these code changes will make `sanoid` skip an update of the cache file, even if it has expired, if the following logic conditions are all met:

- The commands passed to `sanoid` include a `--monitor-*` command; and
- **No** command that takes a snapshot or prunes a snapshot are supplied.

Fixes jimsalterjrs/sanoid#851

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -4,7 +4,7 @@
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
 # project's Git repository at https://github.com/jimsalterjrs/sanoid/blob/master/LICENSE.
 
-$::VERSION = '2.2.0';
+$::VERSION = '2.2.1';
 my $MINIMUM_DEFAULTS_VERSION = 2;
 
 use strict;
@@ -33,6 +33,14 @@ GetOptions(\%args, "verbose", "debug", "cron", "readonly", "quiet",
 if (keys %args < 4) {
 	$args{'cron'} = 1;
 	$args{'verbose'} = 1;
+}
+my $no_need_for_cache_update = 0;
+# Do not update the snapshot cache file if _only_ "--monitor-*" action commands are given (ignore "--verbose", "--configdir" etc)
+if (($args{'monitor-snapshots'} || $args{'monitor-health'} || $args{'monitor-capacity'}) && ! ($args{'cron'} || $args{'force-update'} || $args{'take-snapshots'} || $args{'prune-snapshots'} || $args{'force-prune'})) {
+	# The command combination above must not assert true for any command that takes or prunes snapshots
+	# As long as no snapshots are taken, no conflict with the $forcecacheupdate variable below should occur
+	$no_need_for_cache_update = 1;
+	if ($args{'debug'}) { print "DEBUG: command combo means that the cache file (provided it exists) will not be updated regardless of age.\n"; }
 }
 
 # for compatibility reasons, older versions used hardcoded command paths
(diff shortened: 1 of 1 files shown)


sanoid
```diff
@@ -4,7 +4,7 @@
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
 # project's Git repository at https://github.com/jimsalterjrs/sanoid/blob/master/LICENSE.
 
-$::VERSION = '2.2.0';
+$::VERSION = '2.2.1';
```
phreaker0, line 7: pls don't increase the version numbers, this will be done on release.

sanoid
```diff
@@ -813,31 +821,42 @@ sub getsnaps {
...
-			@rawsnaps = <FH>;
-			close FH;
-
-			open FH, "> $cache" or die 'Could not write to $cache!\n';
-			print FH @rawsnaps;
-			close FH;
-			removelock('sanoid_cacheupdate');
-		} else {
-			if ($args{'verbose'}) { print "INFO: deferring cache update - valid cache update lock held by another sanoid process.\n"; }
+		if ( -f $cache && ! $forcecacheupdate && $no_need_for_cache_update ) {
```
phreaker0, line 824: you still need the lock so reading of the cache file isn't done in the same moment as it's written by another process.

phreaker0 (changes requested):
wouldn't this be bad if there is a stale cache file? Let's assume due misconfiguration sanoid isn't run and snapshot got deleted somehow. In this case the monitoring system wouldn't notice this immediately only after some specified time  where expected snapshots are missing.

jimsalterjrs:
Is this still relevant, @phreaker0 ?

phreaker0:
> Is this still relevant, @phreaker0 ?

yes, but the comments needs to be addressed first.

sanoid
Pajkastare: Sure

sanoid
Pajkastare: Good point, you mean something like `if (checklock('sanoid_cacheupdate'))`? 

I propose another solution, which also solves your initial comment above:

How about instead of ignoring the age of the cache file (as originally discussed in the issue and PR), the allowed age of the cache file is non-persistently increased instead, and no changes to the `getsnaps` function are made? I.e., the default `cacheTTL` value of 900 (15 minutes) is overwritten with something larger such as 18000 (5 hours), if and only if a "monitor-*" command is supplied and no switch that would either create or prune snapshots is active?

The upside is that there are fewer code changes, fewer special cases, and even if some snapshots are deleted outside of sanoid (or the sanoid timer becomes inactive), after a couple hours the "--monitor-*" commands will let Nagios show that there is a problem (instead of erroneously showing that all is fine).

phreaker0: checklock doesn't lock, you will need to put the cache reading between "writelock('sanoid_cacheupdate');" and "removelock('sanoid_cacheupdate');"

But looking at the existing code, reading the cache is also done without locks so there is already an issues which needs to be locked at (sanoid should only be running once not in parallel so this should only be an issue currently with monitor commands).

Your proposed solution sounds good.

Pajkastare: Then I'll make some code changes and commit again "soon".


Just one comment: I'm aware that checklock doesn't lock. But: The usecase that was the reason for the original issue cannot call "writelock"/"removelock", since the "--monitor-snapshots" etc commands are often run by the user "www-data" or "nagios" or something similar depending on Nagios/NRPE setup - Regardless of user, it is likely (and should be) a non-privileged account.

So unless sudo privileges are granted to the user calling "sanoid --monitor-*" by putting something in `/etc/sudoers.d`, or if that user is granted write access to `/run/sanoid` via `chmod`, then sanoid will fail for disk access rights reasons.

Pajkastare: @phreaker0 : I'm not used to pushing changes to git, but I think the final commit only changes the `cacheTTL` value in case of non-snapshot-modifying arguments. As discussed above.

phreaker0 (approved):
looks good, thx

PR #861, opened 2023-11-06
Allow snapshots and bookmarks to be filtered by name

gwollman:
We want to run syncoid with --no-sync-snap but we do not want to consider all available snapshots as suitable incremental sources, certain specially named snapshots.  With the --filter flag, we can specify a regexp that will force syncoid to only "see" matching snapshots.  This is not ideal: if we ever change which snapshots we use, filtering in this way will make it impossible to find a common snapshot or bookmark to use as an incremental source.  But that would require quite a bit more thought to capture all of the different cases.

You probably don't want to take this as-is due to the hackish nature of how it was implemented but I wanted to at least raise the possibility.  (The way we actually are looking to use this is to have only our `daily-*` snapshots synchronized, even if the most recent `daily` is not the most recent snapshot overall.)

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -17,6 +17,7 @@ use Capture::Tiny ':all';
 
 my $mbuffer_size = "16M";
 my $pvoptions = "-p -t -e -r -b";
+my $filter_regexp;
 
 # Blank defaults to use ssh client's default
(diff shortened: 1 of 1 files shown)


phreaker0:
i think this use case will be already covered with #699

PR #872, opened 2023-12-11
fix zfs-get for "--preserve-properties" and tests

Rantherhin:
Some syncoid tests and the "--preserve-properties" flag are failing on ZoL/ZoF devices with the following test errors:
```shell
DEBUG: getting locally set values of properties on syncoid-test-9/src/zvol16...
  zfs get all -s local -H 'syncoid-test-9/src/zvol16'
WARNING: getlocalzfsvalues failed for syncoid-test-9/src/zvol16: cannot open '-s
': dataset does not exist
cannot open 'local': dataset does not exist
cannot open '-H': dataset does not exist
``` 
```shell
+ zfs get recordsize -t filesystem -r syncoid-test-7/dst
cannot open '-t': dataset does not exist
cannot open 'filesystem': dataset does not exist
cannot open '-r': dataset does not exist
``` 
Fixed by moving the property field directly before the dataset for zfs-gets in the affected files. This mimics how zfs-get is used throughout the rest of the project.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1357,9 +1357,9 @@ sub getlocalzfsvalues {
 	if ($debug) { print "DEBUG: getting locally set values of properties on $fs...\n"; }
 	my $mysudocmd;
 	if ($isroot) { $mysudocmd = ''; } else { $mysudocmd = $sudocmd; }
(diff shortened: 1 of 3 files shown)


phreaker0:
thx

jimsalterjrs:
needs rebase

phreaker0:
rebased

PR #882, opened 2024-01-09
escape property key and value pair in case of property preservation

### LINKED ISSUE #852: syncoid doesn't properly escape zfs property values before passing them to the shell
In https://github.com/jimsalterjrs/sanoid/blob/61000c9da2f8762424ba2de8a49abb1b7ec5154d/syncoid#L496, the value of a zfs property is placed in a string and then apparently passed to a shell without quoting.

I noticed this when a user property contained whitespace and this caused the resulting zfs command line to fail; but under unlucky circumstances, or when some zfs properties are controlled by an adversary (e.g. due to `zfs allow userprops`, or because you're working with a pool that was created by someone else), the consequences could be worse.

Instead of the quick-and-dirty `escapeshellparam` function  I recommend you use the established https://metacpan.org/dist/String-ShellQuote/source/ShellQuote.pm for safe shell escaping.

In the longer term, perhaps it would be better to construct the entire command line in an array and pass that to `system()`, not involving the shell at all.

### LINKED ISSUE #862: syncoid with --preserve-properties fails because values aren't escaped
Hi,
In syncoid [L496](https://github.com/jimsalterjrs/sanoid/blob/61000c9da2f8762424ba2de8a49abb1b7ec5154d/syncoid#L496)  there is `$recvoptions .= " -o $key=$value";`
It fails for values with shell active characters like 15years=>3months, something along `escapeshellparam()` is needed.
Keys shouldn't need escaping because from `man zfs`
>  They may contain lowercase letters, numbers,
>     and the following punctuation characters: colon (":"), dash ("-"), period ("."), and underscore ("_")

Perhaps a [test](https://github.com/jimsalterjrs/sanoid/blob/f3d4d309b5a6ccba11512b698041e80a0c8518fa/tests/syncoid/9_preserve_properties/run.sh)  with user properties like net.openoid:var-name should also be added.

phreaker0:
Fixes #852 
Fixes #862

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -493,7 +493,8 @@ sub syncdataset {
 			foreach my $key (keys %properties) {
 				my $value = $properties{$key};
 				if ($debug) { print "DEBUG: will set $key to $value ...\n"; }
-				$recvoptions .= " -o $key=$value";
+				my $pair = escapeshellparam("$key=$value");
+				$recvoptions .= " -o $pair";
 			}
 		} elsif (defined $args{'preserve-recordsize'}) {
 			my $type = getzfsvalue($sourcehost,$sourcefs,$sourceisroot,'type');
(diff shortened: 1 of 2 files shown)


jimsalterjrs:
Needs rebasing

phreaker0:
rebased

PR #883, opened 2024-01-11
update possible zfs send options

### LINKED ISSUE #805: sendoption -b is ignored
First of all thank you for this amazing program!

I was doing playing with `syncoid`, creating and restoring backups, and I found that `zfs send` option `-b` is ignored. 

I added the `b` option in the following line https://github.com/jimsalterjrs/sanoid/blob/ef601eb96b016786af63ea5d7e30c8ad1b5de689/syncoid#L403  and now it works, but I was wondering if this option is not supported for a reason or it's just a bug.

Also, I'm not sure if it should be added in the other calls of `getoptionsline(\@sendoptions, ...)`.

Thank you in advance

phreaker0:
'D' is deprecated and only there for backward compatibility
'V' updates the process title with states
'X' 's' is for excluding datasets and skipping missing snapshots with 'R'
'b' seems to be new

Fixes #805

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -451,7 +451,7 @@ sub syncdataset {
 		# with ZFS on Linux (possibly OpenZFS in general) when setting/unsetting readonly.
 	#my $originaltargetreadonly;
 
-	my $sendoptions = getoptionsline(\@sendoptions, ('D','L','P','R','c','e','h','p','v','w'));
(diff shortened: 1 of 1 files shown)


jimsalterjrs:
Needs rebasing since merging in the in/ex-clude snaps options

phreaker0:
rebased

PR #884, opened 2024-01-15
sanoid.conf: document two options for recursive

dlangille:
zfs and yes are the options, one uses zfs, the other sanoid code


phreaker0:
thx

PR #914, opened 2024-04-20
sanoid #912: sanoid --prune-snapshots performance boost by making --f…

sdettmer:
…orce-prune default (and adding --double-check-prune to get old behavior)

Option dev/sde/double-check-prune - I think the name --no-force-prune is irritating, so it might be renamed to --double-check-prune


phreaker0:
we will use #915

PR #915, opened 2024-04-20
sanoid #912: sanoid --prune-snapshots performance boost by removing u…

sdettmer:
…nneeded iszfsbusy()

Option 3) dev/sde/remove-force-prune - If we think the whole functionality is not needed, best would to remove it and keep things more clean. Maybe this could be done in a later version, when experiences shows, that no one ever used --no-force-prune or --double-check-prune. Doing this later probably is too much, but I work in an industry with 20 years warranty periods so I'm used to think this way :)


phreaker0:
@jimsalterjrs made the necessary changes, this is now ready to be merged

phreaker0 (approved):

lckarssen:
Just adding a reminder that [the Wiki page](https://github.com/jimsalterjrs/sanoid/wiki/Sanoid#flags) needs to be updated as well once this is merged.

PR #916, opened 2024-04-20
syncoid: Add zstdmt compress options

0xFelix:
Add the zstdmt-fast and zstdmt-slow compress options to allow use of multithreading when using zstd compression.


0xFelix:
Ping @phreaker0 @jimsalterjrs

jimsalterjrs:
Thanks for the PR!

PR #959, opened 2024-10-25
Syncoid: Allow hostid as argument

rayzilt:
When using an ZFS pool in a multihost configuration, the hostname can change based on where the pool is imported.
Syncoid creates ephemeral snapshot that contains the hostname from where the command is started. When the pool is failover to another machine, the hostname changes and causes to left over ephemeral snapshots from Syncoid on both sides.

Using --no-sync-snap can be a workaround, but in this situation we like to have ephemeral snapshot of Syncoid.

This PR adds the --hostid argument to specifiy a custom hostname / clustername instead. If not specified it will use the hostname.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -148,7 +148,7 @@ if (length $args{'hostid'}) {
 		pod2usage(2);
 		exit 127;
 	}
-	$hostid = "$args{'hostid'}_";
+	$hostid = "$args{'hostid'}";
 }
 
 # figure out if source and/or target are remote.


jimsalterjrs:
What does the argument you add in your patch do that the existing **--identifier** argument does not? 

```
--identifier=

Adds the given identifier to the snapshot and hold name after "syncoid_" prefix and before the hostname. This enables the use case of reliable replication to multiple targets from the same host. The following chars are allowed: a-z, A-Z, 0-9, _, -, : and . .
```

rayzilt:
Thank you for looking into this. 

I've used the --identifier but noticed that snapshots where still present from the other host after the pool failover. Looking at https://github.com/jimsalterjrs/sanoid/blob/master/syncoid#L1689 it seems that the destroy sub looks at the identifier and hostid.

I'm not syncing to multiple targets, but to two machines that maintain one pool in a multihost configuration.

Maybe I was running into a bug, and responded to soon with this new argument?

rayzilt:
I'll try to investigate why the --identifier argument is not working on my side, as it should be the same functionality as you pointed out. Thanks :-)

jimsalterjrs:
I'm not the one who wrote --identifier IIRC, so it's also possible that **I** misunderstood the way **it** is supposed to work.

If you can confirm that using --identifier doesn't help with your situation, the next step would be expending some skull sweat figuring out whether it makes more sense to implement your proposed --hostid argument in **addition** to --identifier, or whether it makes more sense to simply change how --identifier works.

Personally, I'm having some difficulty coming up with a good reason to separate the two functions out as separate arguments. It seems to me as though if you're already examining things to the degree that using --identifier implies, you should be aware enough of what you're doing to recognize that using the same --identifier in places where it might conflict is a problem to avoid.

@phreaker0 do you have any thoughts on this one?

wfdewith:
For my use case `--identifier` is not sufficient. I'm syncing a pool to an external HDD periodically as a backup. I can connect this external HDD to different hosts to run the same sync, which means that the hostname is not always the same. This means that the sync snapshots for host `foo` are not pruned when I perform the sync from host `bar`. At the moment, I use an UTS namespace to temporarily change the hostname, but that is obviously not an ideal solution.

I'm not sure about the exact use case for `--identifier` but as it currently works, it is useful if you want to distinguish sync snapshots if you have multiple hosts with the same hostname, but different ZFS pools, as opposed to multiple hosts with different hostnames but the same ZFS pool.

In my opinion, the ideal solution to merge these use cases is to have `--identifier` override the hostname part of the sync snapshot name, so that their name becomes `syncoid_<identifier>_<timestamp>` instead of `syncoid_<identifier>_<hostname>_<timestamp>`. In that situation, if you need the hostname in the sync snapshot name, you can add it yourself. Unfortunately, this would be a breaking change for current users of `--identifier`, so it might not be a good way forward.

PR #961, opened 2024-11-05
docs: clarify that scripts are run only if autosnap or autoprune are set

alexgarel:
This PR add a small clarification, although it might sound obvious to some people, it is better to make it explicit.


phreaker0:
thx

alexgarel:
Hello, shan't we merge the branch ?

PR #971, opened 2024-12-18
Add disclaimer if using FreeBSD 14.1

zenodotus280:
I was able to use initiate `syncoid` successfully from a Debian-based host to a fresh installation of FreeBSD with no configuration changes at all. I hadn't even installed mbuffer or pv - I merely started SSH and was able to replicate right away.

After installing sanoid/syncoid on FreeBSD I found that all the warnings no longer applied: `/usr/local/bin` is the default for all relevant executables and the shebangs were correct.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -162,6 +162,7 @@ Install prerequisite software:
 ```bash
 pkg install p5-Config-Inifiles p5-Capture-Tiny pv mbuffer lzop sanoid
 ```
+_No additional steps are required as of FreeBSD 14.1 - the notes below may apply to versions prior to that._
 
 **Additional notes:**
 


jeffsf:
I believe this is from `USES=           shebangfix perl5` and `SHEBANG_FILES=  findoid sanoid sleepymutex syncoid` in the port Makefile. Installing from ports or pkg seems an approach to recommend over direct install from source.

`git blame` suggests that has been the case since 2019 -- https://cgit.freebsd.org/ports/commit/sysutils/sanoid/Makefile?id=56056e7d4bf21a333b4f9df66c7867060359642f

jimsalterjrs:
A lot of folks install from master here rather than from FreeBSD ports, so they still need to do their own localization manually.

PR #980, opened 2025-01-21
Update INSTALL.md

MynaITLabs:
Some exceptions for 

CentOS Stream 9 
* No Powertools

AlmaLinux 9.5 
* No Powertools
* Not in root path /usr/local/sbin and /usr/local/bin


MynaITLabs:
Not sure if this is the best method for suggesting simple edits or not.

MynaITLabs:
Ok, looks like a known issue.  The package is now called 'crb' instead.
https://unix.stackexchange.com/questions/678777/how-to-get-epel-release-epel-release-next-powertools-in-centos-stream-9

MynaITLabs:
Retracted PR. Working on a few more changes to request.

Issue #552, opened 2020-05-27
Feature request: Ability to change snapshot name format


darkpixel:
I would love the ability to change the snapshot name format.

It's a minor nitpick, but if you put these two snapshot names in to your terminal:
```
tank/officeshare@zfs-auto-snap_04-2020-03-01-1600
tank/officeshare@syncoid_usrbgofnas01_2020-05-01:20:42:59
```
...and then try to double-click each one to select it, the snapshot created by zfs-auto-snapshot automatically selects the entire name.  The snapshot by sanoid only selects ```tank/officeshare@syncoid_usrbgofnas01_2020-05-01``` because it runs into a colon.

I would love the ability to change the snapshot name format so date and time are separated by an underscore and the the individual fields are separated by dashes.

Yvan-Masson:
Hi,

This feature would be essential for my use case: I need to share snapshots via SMB (Samba) to Windows computers, which is currently impossible because the colon is a forbidden character. Currently, snapshot name is mangled by Samba, which makes it meaningless.

Regards,
Yvan

phreaker0:
@Yvan-Masson you can use the samba catia module to remap the illegal chars, see: https://github.com/jimsalterjrs/sanoid/issues/345#issuecomment-459660254

Yvan-Masson:
@phreaker0 Thanks for the tip! I have also just seen samba shadow_copy2 module which might not be trivial to setup but could do the job (https://github.com/jimsalterjrs/sanoid/issues/5#issuecomment-730660664).

reefland:
It would be helpful to alter snapshot name to remove the seconds from the snapshot name.  I have TrueNAS pulling Sanoid snapshots and TrueNAS does not allow seconds as a variable in schema name. So I have to hard code that as ":00", but often Sanoid second is at 01 which is annoying because we don't need seconds in a snapshot name.

jimsalterjrs:
I'm not sure what you mean by "TrueNAS pulling Sanoid snapshots." Can you elaborate?

reefland:
Jim, be glad to explain it...

TrueNAS (previously known as FreeNAS) has the ability to PUSH out to other servers and/or PULL in from other servers;  ZFS snapshots using its built in "Replication Tasks".  Behind the scenes the SSH session happens and the normal ZFS send / receive happens.

[...]

PR #988, opened 2025-02-21
Ability to change snapshot name format

kennypm:
Allows for setting custom date format and snapshot name format per template, as discussed in #552.


ArthurKamalov:
@kennypm hey, 
I am also looking for that feature. Any reasons why did you close it?

kennypm:
name formatting is working as expected, but broke the rest of it such that
all snap types are being taken every time sanoid is run, regardless of what
the cache and snaps say. I didn't test thoroughly enough before submitting.
once I fix I will resubmit. contributions are welcome

PR #989, opened 2025-02-22
Ability to change snapshot name format

kennypm:
This should address everyone's requests from #552, succinctly summed up by @ams-tschoening:

> * custom prefix and suffix
>   * e.g. replace `autosnap` with some custom wording
>   * e.g. move labels like `frequently` to the front
> * custom format of timestamps
>   * e.g. to remove `:`
>   * e.g. to make format overall shorter, e.g. `@autosnap_20230303T203610Z_weekly`
> * maintain formats in the config file
>   * templates + overrides per dataset would be great
>   * I put config files those under version control mostly

The new properties can be overridden for child datasets if the parent uses `recursive = yes`. I've tested a few different configurations and everything seems to be working as expected. The first commit here covers everything except reordering the substrings with a user-defined template.

I separated that into a second commit since it requires more invasive changes to `getsnaps()`. The old behavior expects snap names to start with `autosnap` and end with a `*ly` type. The new behavior recognizes sanoid autosnaps as long as the name contains both a `*ly` type and whatever identifier string the dataset is configured to use, which defaults to `autosnap`.

There is currently no tracking of old identifier labels after they've been changed, but the only person who mentioned that possibility said they didn't need it.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -17,6 +17,7 @@ use Getopt::Long qw(:config auto_version auto_help);
 use Pod::Usage;       # pod2usage
 use Time::Local;      # to parse dates in reverse
 use Capture::Tiny ':all';
+use POSIX 'strftime';
 
 my %args = (
 	"configdir" => "/etc/sanoid",
@@ -638,7 +639,8 @@ sub take_snapshots {
 			my @snapshots;
 
 			foreach my $type (@types) {
-				my $snapname = "autosnap_$datestamp{'sortable'}_$type";
+				my $sortable = strftime($config{$dataset}{'datestamp_format'}, localtime($datestamp{'unix_time'}));
+				my $snapname = "$config{$dataset}{'prefix'}_${sortable}_$type";
 				push(@snapshots, $snapname);
 			}
 
@@ -947,7 +949,7 @@ sub getsnaps {
 		# avoid pissing off use warnings
 		if (defined $snapname) {
 			my ($snaptype) = ($snapname =~ m/.*_(\w*ly)/);
-			if ($snapname =~ /^autosnap/) {
+			if ($snapname =~ /^$config{$fs}{'prefix'}/) {
 				$snaps{$fs}{$snapname}{'ctime'}=$snapdate;
 				$snaps{$fs}{$snapname}{'type'}=$snaptype;
 			}
@@ -1184,16 +1186,15 @@ sub init {
 
 sub get_date {
 	my %datestamp;
-	($datestamp{'sec'},$datestamp{'min'},$datestamp{'hour'},$datestamp{'mday'},$datestamp{'mon'},$datestamp{'year'},$datestamp{'wday'},$datestamp{'yday'},$datestamp{'isdst'}) = localtime(time);
+	$datestamp{'unix_time'} = time();
+	($datestamp{'sec'},$datestamp{'min'},$datestamp{'hour'},$datestamp{'mday'},$datestamp{'mon'},$datestamp{'year'},$datestamp{'wday'},$datestamp{'yday'},$datestamp{'isdst'}) = localtime($datestamp{'unix_time'});
+
 	$datestamp{'year'} += 1900;
-	$datestamp{'unix_time'} = (((((((($datestamp{'year'} - 1971) * 365) + $datestamp{'yday'}) * 24) + $datestamp{'hour'}) * 60) + $datestamp{'min'}) * 60) + $datestamp{'sec'};
 	$datestamp{'sec'} = sprintf ("%02u", $datestamp{'sec'});
 	$datestamp{'min'} = sprintf ("%02u", $datestamp{'min'});
 	$datestamp{'hour'} = sprintf ("%02u", $datestamp{'hour'});
 	$datestamp{'mday'} = sprintf ("%02u", $datestamp{'mday'});
 	$datestamp{'mon'} = sprintf ("%02u", ($datestamp{'mon'} + 1));
-	$datestamp{'noseconds'} = "$datestamp{'year'}-$datestamp{'mon'}-$datestamp{'mday'}_$datestamp{'hour'}:$datestamp{'min'}";
-	$datestamp{'sortable'} = "$datestamp{'noseconds'}:$datestamp{'sec'}";
 	return %datestamp;
 }
 
--- a/sanoid
+++ b/sanoid
@@ -640,7 +640,10 @@ sub take_snapshots {
 
 			foreach my $type (@types) {
 				my $sortable = strftime($config{$dataset}{'datestamp_format'}, localtime($datestamp{'unix_time'}));
-				my $snapname = "$config{$dataset}{'prefix'}_${sortable}_$type";
+				my $snapname = $config{$dataset}{'snapname_format'};
+				$snapname =~ s/IDENTIFIER/$config{$dataset}{'identifier'}/g;
+				$snapname =~ s/DATE/$sortable/g;
+				$snapname =~ s/TYPE/$type/g;
 				push(@snapshots, $snapname);
 			}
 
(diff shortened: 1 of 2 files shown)


beren12:
Can't wait! this should really help with window's previous versions and samba

phreaker0:
@kennypm this looks really good. It should preserve the default behaviour nicely and everyone can adapt the format in a very flexible way. I still need to run the tests in a vm to make sure it works and will report back.
@jimsalterjrs i guess this works for you too if everything checks out?

jimsalterjrs:
Well, yes, but I do have to wonder if it's quite the best way forward. Wouldn't it be better to abandon "recognizing" snapshots by name format at all, and instead begin recognizing them either by the original, default naming format **or** by  setting a custom ZFS property when sanoid (or syncoid!) takes a snapshot?

That way we can make our sweeping change now, Sanoid will not only let people change the naming format but **will keep working on all existing snapshots even after a naming format change**, etc.

Eventually, we can consider dropping name-based detection support at all, and rely purely on the custom ZFS property (or properties, there's probably a *lot* we could be doing with those if we put our minds to it).

What do y'all think?

phreaker0:
@kennypm  Running the tests I get the following on the first run of sanoid:

```
+ ../../sanoid --cron --verbose
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value in regexp compilation at ../../sanoid line 899.
Use of uninitialized value $path in hash element at ../../sanoid line 787.
Use of uninitialized value $path in hash element at ../../sanoid line 825.
Use of uninitialized value $path in hash element at ../../sanoid line 825.
Use of uninitialized value $path in hash element at ../../sanoid line 829.
Use of uninitialized value $path in hash element at ../../sanoid line 834.
Use of uninitialized value $path in hash element at ../../sanoid line 840.
INFO: taking snapshots...
taking snapshot sanoid-test-1@autosnap_2017-01-01_00:00:00_monthly
taking snapshot sanoid-test-1@autosnap_2017-01-01_00:00:00_daily
taking snapshot sanoid-test-1@autosnap_2017-01-01_00:00:00_hourly
```

kennypm:
When I updated to v2.3.0, my patch broke and I never bothered rebasing the branch due to the lack of movement here at that point. I'll gladly do the work of fixing it up if we can come to consensus on the underlying mechanism as @jimsalterjrs discussed above. I remember being a little surprised as I was developing the feature that it was relying on pattern matching of snapshot names, so I guess my vote would be for a ZFS property for the sake of robustness.

phreaker0:
> Well, yes, but I do have to wonder if it's quite the best way forward. Wouldn't it be better to abandon "recognizing" snapshots by name format at all, and instead begin recognizing them either by the original, default naming format **or** by setting a custom ZFS property when sanoid (or syncoid!) takes a snapshot?

Even if we use a custom property for more flexibility in the future it would still be very helpful to define the snapshot naming format for human readable snapshot lists in the filesystem, for samba's shadow_copy2 module, ... .

> Eventually, we can consider dropping name-based detection support at all, and rely purely on the custom ZFS property (or properties, there's probably a _lot_ we could be doing with those if we put our minds to it).

I think dropping name based detection support is not a good idea because it will probably create some issues:

1. by default syncing datasets will omit properties, so it's not guaranteed sanoid will detect a replicated pool any more despite the snapshots being there 
2. you can't easily move from other snapshot tools/custom scripts to sanoid

jumbi77:
+1 Looking forward to this patch!

PR #996, opened 2025-03-31
syncoid: fix directtimeout in directmbuffer mode

jjakob:
If --insecure-direct-connection contained 4 parts (including the ',mbuffer' at the end), the 3rd part (timeout) was silently ignored and left at the default 60s.

Do not ignore the timeout part even in directmbuffer mode.


phreaker0:
thx

Issue #972, opened 2024-12-18
[syncoid] Feature request - separate argument for ZFS properties


Peter2121:
I have a problem when I try to send a dataset with many children to a backup host.

To avoid mounting the datasets received on a backup host, I need to set ZFS property `canmount` to `noauto` (using `-u` does not help as the dataset will be mounted on reboot). So, I use syncoid with argument `--recvoptions='-u -o canmount=noauto'`. The problem is that some datasets must be synced with their properties (for example, to avoid loosing the information about mountpoints), so `--preserve-properties` argument must be used as well. Unfortunately, some datasets have ZFS property `canmount` set to something. For example, in FreeBSD the dataset `/usr` has `canmount` set to `off`. `zfs receive` command failed in such case: `property 'canmount' specified multiple times`.

I think that syncoid needs a separate argument to manage zfs properties to set by `zfs receive`. For example, something like `--recvprop='canmount:noauto'` (with a possibility to be specified multiple times). The properties set at this manner should override the properties got from the source dataset (for example, in case of `/usr`, the destination dataset will have `canmount=noauto`).

I would be happy to help with this, but I'm zero in Perl :(

...But I'm ready to help with testing/debugging ;)

Peter2121:
...And there is the same problem on restore - at the same time I need to use `--preserve-properties` (for example, to restore the mountpoint) and I need to pass `--recvoptions='-o canmount=on'` to mount the dataset immediately or on reboot.

jimsalterjrs:
> I think that syncoid needs a separate argument to manage zfs properties to set by `zfs receive`.

How do you picture this working? Forget Perl, simply in terms of the OpenZFS utilities themselves, on the command line. 

To the best of my knowledge, you've already got full access to the **entire** gamut of what's possible with replication by way of the --sendoptions and --recvoptions arguments. I don't know what else you could actually do, even at the command line itself.

Peter2121:
I have the patched version already, it still needs more tests. I'll send a PR once I consider it's OK.

PR #997, opened 2025-04-10
Implement --restore-properties and --target-properties

Peter2121:
Following issue #972
The implemented features permit to define the properties for destination dataset without loosing the initial properties that can be restored later.
For example, it is now possible to safely backup and restore datasets with different 'mountpoint' and 'canmount' options. It could be useful to backup the whole 'zroot' dataset to another host and then restore it, without manually changing any option on backup and on restore.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -27,7 +27,7 @@ GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsn
                    "no-clone-handling", "no-privilege-elevation", "force-delete", "no-rollback", "create-bookmark", "use-hold",
                    "pv-options=s" => \$pvoptions, "keep-sync-snap", "preserve-recordsize", "mbuffer-size=s" => \$mbuffer_size,
                    "delete-target-snapshots", "insecure-direct-connection=s", "preserve-properties",
-                   "include-snaps=s@", "exclude-snaps=s@", "exclude-datasets=s@")
+                   "include-snaps=s@", "exclude-snaps=s@", "exclude-datasets=s@", "target-properties=s@", "restore-properties")
                    or pod2usage(2);
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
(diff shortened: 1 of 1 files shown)


ifazk:
Adding two additional flags seem a bit heavy. I think it would be easier to just add an option to add additional properties to the property blacklist.

Uses for `--target-properties` is already handled by `recvoptions`. By the time users needs to use `--target-properties`, they are already pretty advanced users, so I'd expect them to be familiar with `recvoptions`.

Peter2121:
The most interesting part of my approach is backing up the properties to change and restoring them later. I don't see any way to implement this without the proposed flags.

ifazk:
Sorry. I totally misunderstood what `--restore-properties` was doing, that's not a bad idea.

phreaker0:
@Peter2121 I only had a brief look so far:

- "preserve-properties" functionality is changed but this should behave like before, there should be a new flag for this feature
- getlocalzfsvalues is now always called, even if not needed

can you explain the feature in more detail, I don't understand it fully :-)

Peter2121:
Sorry for delay.

> "preserve-properties" functionality is changed but this should behave like before, there should be a new flag for this feature

The functionality of "preserve-properties" remains unchanged in case of absence of _new_ option "target-properties" (checked at [line 984](https://github.com/Peter2121/sanoid/blob/ebcedbc89b58cb3fbe5bd97bcf53fd7ec6fc99f1/syncoid#L984)). So, I don't see any compatibility problem with the existing scripts. And it _must_ be changed in case of presence of "target-properties". So, if anyone adds "target-properties" he must review the functionality of "preserve-properties" in such case.


> "getlocalzfsvalues" is now always called, even if not needed

The result of "getlocalzfsvalues" is needed in different code parts. So IMHO it would be better to have the result _always_ rather than put several "if" and call this function just sometimes. This call does not seem to be too hard to add complexity in the code to avoid it. If you think different - it is surely possible to put in in "if" block checking presence of related options.

Peter2121:
Globally, the idea is that when we copy a snapshot to a different system - some dataset properties must be changed to avoid breaking changes on the destination system. For example, if you sync a snapshot of a root dataset - it must not be mounted. You can imagine another cases like datasets for jails with the same names synced from different hosts to a backup host. BUT if you just overwrite the dataset properties during the sync - you will loose the 'old' values completely. So, if you want to restore the dataset on the original host - you need a source to take the data from. That's why I added "target-properties" argument. If it is present - the properties on destination will be set but the local properties of the dataset on source host will be saved on destination with "syncoid:" prefix. Then, during sync with "restore-properties" the saved properties can be restored on the initial host.

PR #1009, opened 2025-06-17
get snpshots from configured list of datasets only

bjoern-r:
query the existing snapshots only from the configured datasets to avoid spinning up disks that are not in the config file

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -883,7 +883,8 @@ sub getsnaps {
 					print "INFO: cache expired - updating from zfs list.\n";
 				}
 			}
-			open FH, "$zfs get -Hrpt snapshot creation |";
+			# just get snapshots from configured datasets to not spin up the disks
+			open FH, "$zfs get -Hrpt snapshot creation ".join(" ",keys %$config)." |";
 			@rawsnaps = <FH>;
 			close FH;
 


phreaker0:
interesting but I can think of two problems:

1.  there is a command character limit, if there are many datasets this may lead to a situation where the command fails to be executable, but I think this limit is quite high but this could be handled with some extra code to split the execution calls but I don't know if this is needed
2. if one uses large cache expire times and the dataset configuration is changed it will appear to sanoid that there are no snapshots for the new configured datasets and will create ones. But this will lead only to one extra set of snapshots in the worst case and I don't think this an issue.

bjoern-r:
1) i also thought of this possible limitation, however my perl skills are somewhat limited and i did not want to introduce bad code so maybe this could be handled by someone with better knowledge of perl and sanoid in particular.
2) would this extra snapshot be immediately been purged by a sanoid-purge run because it is in the wrong interval?

phreaker0:
1) I also don't know if older zfs commands support to specify multiple pools so we need a fallback any way. Can you add an error check for the new command, if it fails it should run the old one without specifying any pools. This way we would be backwards compatible and it would still work if we exceed the command lenght limit.
2) no

eigenphase:
Thanks @bjoern-r , suffering from this too and hope it makes it in!!

Re 1: I don't think this is a practical limit. It's literally megabytes in modern systems:

```
# getconf ARG_MAX
2097152
```

What's probably the easiest it to query this, then compare the resulting command and if it's longer, just print a warning and use "zfs list", instead of your expanded command.
I don't think it would happen to anyone in practice though.

eigenphase:
@bjoern-r  any chance you’re motivated to implement the proposed [changes](https://github.com/jimsalterjrs/sanoid/pull/1009#issuecomment-3051227998)? 
Would be so great to get this PR in

bjoern-r:
i will try to work on this when i find the time, however as already meantioned my perl skills are limited but i will try my best.

jimsalterjrs:
@phreaker0 does this look merge ready to you? It's an interesting feature add but strikes me as pretty risky in terms of the potential to introduce breaking bugs.

bjoern-r:
i added a fallback to the old command in case the more specific command fails.

eigenphase:
@phreaker0 would you be open to check Bjoern's change and merging if ok? 

Works well for me but I'd really love to see it in the official branch

phreaker0:
The error handling didn't work and so the fallback was never used. Fixed and tested it, also made some code style changes to be more consistent. This is now ready to go. Thx.

eigenphase:
Amazing, thank you so much everyone!

Issue #1012, opened 2025-06-28
syncoid feature request: replicate source inherited properties for child datasets that are not inherited at destination


evan314159:
syncoid --preserve-properties preserves locally set properties but not inherited properties that would be not be inherited at the destination.  Example:

tank sets acltype=posix (default for TrueNAS pools)
tank/home sets acltype=nfsv4
tank/home/evan inherits acltype=nfsv4
tank/home/evan/music inherits acltype=nfsv4

backup sets acltype=posix (default for TrueNAS pools)

syncoid --recursive --preserve-properties tank/home/evan backup/evan

replicates the datasets but with acltype inherited from backup = posix, and not acltype=nfsv4 as was inherited from tank/home.

The desired result in this case would be to set acltype=nfsv4 locally on backup/evan and inherit acltype on backup/evan/music.  It would also need to preserve some settings that are intended to be different, for example setting copies=2 on backup.

ghost:
Created PR #1013 to implement this functionality

PR #1013, opened 2025-06-28
syncoid: add capability to preserve inherited properties

ghost:
syncoid --preserve-properties preserves locally set properties but not inherited properties that would be not be inherited at the destination. Example:

```
tank                     local     acltype=posix (default for TrueNAS pools)
tank/home                local     acltype=nfsv4
tank/home/evan inherits  local     acltype=nfsv4
tank/home/evan/Documents inherited acltype=nfsv4
backup                   local     acltype=posix (default for TrueNAS pools)
```
syncoid --recursive --preserve-properties tank/home/evan backup/evan

replicates the datasets but with acltype inherited from backup = posix, and not acltype=nfsv4 as was inherited from tank/home.

The desired result in this case would be to set acltype=nfsv4 locally on backup/evan and inherit acltype on backup/evan/music. It would also need to preserve some settings that are intended to be different, for example setting copies=2 on backup, or excluding mountpoint.

This PR implements a new command line option --preserve-inherited-properties that preserves properties where they were inherited in their source location but would not otherwise be inherited in their destination.

Real output in my enviroment after running syncoid syncoid --recursive --preserve-inherited-properties --recvoptions="u o readonly=on x mountpoint" tank backup

> zfs get acltype,mountpoint,readonly -r backup -t fs

```
NAME                                      PROPERTY    VALUE                                                    SOURCE
backup                                    acltype     posix                                                    local
backup                                    mountpoint  /mnt/backup                                              local
backup                                    readonly    off                                                      default
backup/tank                               acltype     posix                                                    local
backup/tank                               mountpoint  /mnt/backup/tank                               inherited from backup
backup/tank                               readonly    on                                                       local
backup/tank/home                          acltype     nfsv4                                                    local
backup/tank/home                          mountpoint  /mnt/backup/tank/home                          inherited from backup
backup/tank/home                          readonly    on                                                       local
backup/tank/home/evan                     acltype     nfsv4                                                    inherited from backup/tank/home
backup/tank/home/evan                     mountpoint  /mnt/backup/tank/home/evan                     inherited from backup
backup/tank/home/evan                     readonly    on                                                       local
backup/tank/home/evan/Documents           acltype     nfsv4                                                    inherited from backup/tank/home
backup/tank/home/evan/Documents           mountpoint  /mnt/backup/tank/home/evan/Documents           inherited from backup
backup/tank/home/evan/Documents           readonly    on                                                       local

```
This change:
- adds --preserve-inherited-properties command line option with --help
- implicitly sets --preserve-properties if --preserve-inherited-properties is set
- renames getlocalzfsvalues to getzfspropertiestopreserve to reflect its purpose
- returns local and inherited properties if --preserve-inherited-properties is set, otherwise local properties only
- skips propeties that are set or excluded in recvoptions
- skips properties that would be inherited correctly at the destination unless they were locally set at the source
- adds keylocation to blacklist (along with other encryption-related properties)

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -26,11 +26,16 @@ GetOptions(\%args, "no-command-checks", "monitor-version", "compress=s", "dumpsn
                    "debug", "quiet", "no-stream", "no-sync-snap", "no-resume", "exclude=s@", "skip-parent", "identifier=s",
                    "no-clone-handling", "no-privilege-elevation", "force-delete", "no-rollback", "create-bookmark", "use-hold",
                    "pv-options=s" => \$pvoptions, "keep-sync-snap", "preserve-recordsize", "mbuffer-size=s" => \$mbuffer_size,
-                   "delete-target-snapshots", "insecure-direct-connection=s", "preserve-properties",
+                   "delete-target-snapshots", "insecure-direct-connection=s", "preserve-properties", "preserve-inherited-properties",
                    "include-snaps=s@", "exclude-snaps=s@", "exclude-datasets=s@")
                    or pod2usage(2);
 
 my %compressargs = %{compressargset($args{'compress'} || 'default')}; # Can't be done with GetOptions arg, as default still needs to be set
+
+# --preserve-inherited-properties implies --preserve-properties
+if (defined $args{'preserve-inherited-properties'}) {
+	$args{'preserve-properties'} = 1;
+}
 
 if (defined($args{'exclude'})) {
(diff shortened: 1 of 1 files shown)


ghost:
Resolves issue #1012

phreaker0:
keylocation shouldn't be added to the blacklist as it's a writeable property which can be preserve. Blacklist is for properties which can't be written.

PR #1039, opened 2025-09-16
feat: warn if template can't be found

### LINKED ISSUE #1038: sanoid should crash if it can't find the template requested via use_template
Sort of related to #938 but I'm requesting sanoid to crash, not to warn.

You see, I've been having issue with sanoid for quite some time and never seems to figure out what part of the config I was doing wrong. My issue was that `var/docker` was getting snapshot like crazy even though I don't care at all about that as I just use docker bound volumes. So I just wanted to recursively NOT do snapshots on the docker pools.

Turns out my error was that instead of writing `use_template = docker` I had been writing `use_template = "docker"`.

If you write unexpected keys in the config sanoid will crash with for example `FATAL: cannot load /etc/sanoid/sanoid.conf - please create a valid local config file before running sanoid! at /usr/sbin/sanoid line 874.`. So there is definitely some amount of config checking going on.

I believe **strongly** that if the user requests a template that can't be found sanoid should crash. It is a reasonable expectation of the user given that there already is some config checking.

thiswillbeyourgithub:
Signed-off-by: thiswillbeyourgithub <26625900+thiswillbeyourgithub@users.noreply.github.com>

Fixes #1038

Hi,

This 3 line changes adds an important safety net: if the template cannot be found we crash instead of continuing as usual. In my case I had been trying to use template `"docker"` instead of `docker` for a while and never figured it out until today.

I believe if a template is declared and never used we should also crash.

I can't perl so I asked an LLM and then tested on my end.

edit:

I'm adding a second commit to crash if there are unused templates.

### DIFF
--- a/sanoid
+++ b/sanoid
@@ -1020,6 +1020,12 @@ sub init {
 				$rawtemplate =~ s/\s+$//g;
 
 				my $template = 'template_'.$rawtemplate;
+
+                # Check if template exists
+                if (!exists $ini{$template}) {
+                    die "FATAL ERROR: Template '$rawtemplate' referenced in section [$section] does not exist in $conf_file.\n";
+                }
+
 				foreach my $key (keys %{$ini{$template}}) {
 					if ($key =~ /template|recursive/) {
 						warn "ignored key '$key' from '$rawtemplate' template.\n";
--- a/sanoid
+++ b/sanoid
@@ -1021,9 +1021,9 @@ sub init {
 
 				my $template = 'template_'.$rawtemplate;
 
-                # Check if template exists
+                # Check if template exists - warn otherwise
                 if (!exists $ini{$template}) {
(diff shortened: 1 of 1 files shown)


phreaker0:
> This 3 line changes adds an important safety net: if the template cannot be found we crash instead of continuing as usual. In my case I had been trying to use template `"docker"` instead of `docker` for a while and never figured it out until today.
> 
> I believe if a template is declared and never used we should also crash.

It shouldn't crash, but it should output a warning like with the other configuration mistakes. Too many snapshots in an error case is better than no snapshots.

> I'm adding a second commit to crash if there are unused templates.

this needs to be removed. Unused templates aren't errors.

jimsalterjrs:
There is no way in *hell* we're warning, let alone crashing, simply because unused templates are available.

thiswillbeyourgithub:
Your call. I made those changes. Is that satisfactory?

PR #1040, opened 2025-09-21
fix: propagate send failures in pipeline / pipefail

msladek:
**Problem**
`syncoid` could report success when `zfs send` failed inside the `$synccmd` pipeline (e.g., encrypted source with unloaded key). Because `zfs receive` exited cleanly, the shell returned `0`. This caused further issues e.g. `--use-hold` to release holds despite no data being replicated.

**Change**
Enable shell `pipefail` (if the shell supports it) for the assembled send/receive pipeline so any stage failure causes a non-zero exit. No behavioural change on older shells lacking `pipefail`.

Since the [POSIX.1-2024](https://sortix.org/blog/posix-2024) specification added support for `pipefail`, it's now available just about everywhere, including dash.

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -1750,7 +1750,7 @@ sub buildsynccmd {
 
 		$synccmd .= escapeshellparam($remotecmd);
 	}
-	return $synccmd;
+	return "(set -o pipefail 2> /dev/null) && set -o pipefail; $synccmd";
 }
 
 sub pruneoldsyncsnaps {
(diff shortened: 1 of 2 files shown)


phreaker0:
@msladek did you test with ssh replication?

msladek:
@phreaker0 thanks for reviewing
yes, I run this code daily in my own environment using ssh replication. I actually ran into this bug with ssh, where unloaded keys on the source didn't cause the `syncoid` command to fail.

for the automated test `tests/syncoid/015_send_pipefail/run.sh` I kept it local. the bug is reproducible this way and setting up a full SSH environment in CI felt complex and out of scope for this specific fix. The local test confirms the pipeline exit logic works as expected.

PR #1054, opened 2025-12-11
Remove unused forcerecv variable

ifazk:
Small cleanup. The `$forcedrecv` parameter isn't used anywhere. The `-F` flag is set in `sub runsynccmd`.


phreaker0:
@ifazk thx

PR #1061, opened 2026-01-01
Change shebang to use '#!/usr/bin/env NAME' for portability

tomhoover:
Use the first (perl|bash) found in the user's path, rather than hardcoding
/usr/bin/perl and /bin/bash. This allows for the use of a version of (perl|bash)
of the user's choice, rather than requiring the use of the 'system' installed
version. Multiple installed versions are common when using `homebrew` to install
newer versions of (perl|bash) on MacOS or Linux. This is also needed when using
a version manager such as `asdf` or `mise` to manage multiple language runtime
versions.

### DIFF
--- a/findoid
+++ b/findoid
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 # this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
--- a/sanoid
+++ b/sanoid
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 # this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this
--- a/sleepymutex
+++ b/sleepymutex
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # this is just a cheap way to trigger mutex-based checks for process activity.
 #
--- a/syncoid
+++ b/syncoid
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 
 # this software is licensed for use under the Free Software Foundation's GPL v3.0 license, as retrieved
 # from http://www.gnu.org/licenses/gpl-3.0.html on 2014-11-17.  A copy should also be available in this


phreaker0:
@jimsalterjrs whats your opinion on this? I can find arguments for both sides and I would just leave it as is.

tomhoover:
> I can find arguments for both sides and I would just leave it as is.

@phreaker0, I am genuinely curious as to what the argument against the change would be? I'm not aware of any os (that would be using sanoid) that doesn't support the syntax. If there are incompatibilities out there, please let me know so I can stop using `#!/usr/bin/env perl` in other scripts. Thanks!

jimsalterjrs:
@tomhoover sure, most distros support the syntax just fine... but the necessary environment variable is generally NOT set in the cron shell environment, leading to some really irritating Google-fu just to try to get a cron job to work.

I *promise* you it's easier for you to figure out how to edit a shebang than it is for every Linux user out there to have to figure out how to troubleshoot their cron shell variables and change them to a non-default state. 

This is not our first, second, third, or I'm pretty sure even *fourth* time that a FreeBSD user has confidently suggested this change. The first time, I implemented it and tested in dev. It worked fine when invoked manually, and so I thought the change was good... but then my cron jobs failed to fire overnight.

I spent some time figuring out why the hell it worked when run manually but not from cron, and then once I had figured out what the problem was, it was irritating as hell looking up how to *solve* the problem. I realized this was MUCH bigger a faff than simply changing a shebang if it doesn't suit you, so here we are.

> This allows for the use of a version of (perl|bash) of the user's choice, rather than requiring the use of the 'system' installed version.

So does specifying the interpreter directly on the command line. The shebang is only used if you don't specify an interpreter. If you want to specify one, you can do so on the command line... up to and including `env perl /usr/local/bin/sanoid`.

````
root@elden:/# env perl /usr/local/bin/sanoid
INFO: taking snapshots...
INFO: pruning snapshots...
````

Also, I'm pretty sure the FreeBSD port is already modifying the shebang to `#!/usr/bin/env perl` for you, though it's been quite a minute since I looked closely at the FreeBSD port directly. If you're not using that, maybe give that a shot. If you are using that, and the port maintainer isn't making that change to the shebang, reach out to *them* and ask them to make that change--because it's a mildly nice change for FreeBSD users, but a *terrible* one for Linux users.

Finally, if *none* of the above suits you... just make yourself a wrapper script named sanoid, drop it higher in your path than the real sanoid you're using, and make that wrapper script call the real script with the specified interpreter you want. Now when you just type "sanoid" you actually get `/usr/bin/env perl sanoid` just like you want, without screwing up anybody else's system *or* needing anybody else's help, *and* updating the real version of sanoid from ports, directly from master, etc doesn't require you to change anything after the update.

tomhoover:
@jimsalterjrs Thanks for the detailed reply. I had no problem working around getting sanoid/syncoid running (I've been using them for years), but thought (incorrectly) that changing the shebang would fix a perceived issue. I now understand the other issues that change would cause. Thanks for the great utilities!

PR #1065, opened 2026-01-06
 Fix syncoid incremental base detection when `--exclude-snaps` hides the common base snapshot.

pgassmann:
**Summary**

This fixes an issue where syncoid fails to update the target after a filter was added that matches the latest snapshot on the destination. e.g. previously synced latest hourly snapshot, then exclude hourly snapshots.

```bash
syncoid --recursive --no-sync-snap --no-rollback --use-hold \
    --exclude-snaps='_frequently$' --exclude-snaps='_hourly$' \
    --sendoptions="LcV" \
    root@source:dataset target-dataset
```

- **Problem (repro, minimal)**
  1) Create snapshot: `src@hourly1`.
  2) Sync without filters to `dst` (dst gets only `hourly1`).
  3) Create new snaps: `src@daily1` (to send) and `src@hourly2` (to exclude).
  4) Run with `--exclude-snaps='hourly' --no-sync-snap --no-rollback --use-hold`.
  **Expected:** reuse excluded `hourly1` as the incremental base, send `daily1`, skip `hourly2`.
  **Actual before fix:** excluded `hourly1` was filtered out when choosing the base, so no matching snapshot was found and sync aborted.
  See `tests/syncoid/011_exclude_base_snapshot/run.sh` for the executable repro.

- **Code changes (vs origin/master)**
  - Snapshot listing now supports explicit filtered vs unfiltered views. `getsnaps`/`getsnapsfallback` take `applyfilters` (default on) so we build:
    - Filtered maps for send/prune: `%snaps` plus `%targetsnaps` merged into `%snaps` (respects `--exclude-snaps/--include-snaps`).
    - Unfiltered maps for matching/bookmark fallback: `%snaps_unfiltered` plus `%targetsnaps_unfiltered` merged into `%snaps_unfiltered` (ignores filters so excluded bases remain visible by GUID).
  - `getmatchingsnapshot` consumes the provided unfiltered map (no globals) so an excluded base can still be matched by GUID on the target.
  - Base selection for `--no-sync-snap` uses the assembled snapshot map (`\%snaps`) rather than rebuilding from globals, keeping filtered/unfiltered views consistent through the sync path.
  - Test harness tweaks: initial sync sets holds, and data writes go through the mountpoint so ZFS sees deltas. These affect only the regression test setup.

- **Tests**
  - `tests/syncoid/011_exclude_base_snapshot/run.sh` exercises the filtered-base regression; expect failure on unpatched syncoid (“destination … modified”) and success with this change.
  - Existing filtering coverage remains in `tests/syncoid/010_filter_snaps`.
  - Not re-run in this review.

Disclaimer: This update was created with the help of AI tools, using opencode with GPT-5.1 Codex Max by OpenAI

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -586,7 +586,8 @@ sub syncdataset {
 		my $bookmark = 0;
 		my $bookmarkcreation = 0;
 
-		$matchingsnap = getmatchingsnapshot($sourcefs, $targetfs, \%snaps_unfiltered);
+		my $matchingsnapcreation = $snaps_unfiltered{'source'}{$newsyncsnap}{'creation'};
+		$matchingsnap = getmatchingsnapshot($sourcefs, $targetfs, \%snaps_unfiltered, $matchingsnapcreation);
 		if (! $matchingsnap) {
 			# no matching snapshots, check for bookmarks as fallback
 			my %bookmarks = getbookmarks($sourcehost,$sourcefs,$sourceisroot);
@@ -1740,8 +1741,9 @@ sub pruneoldsyncsnaps {
 	}
 
 sub getmatchingsnapshot {
-	my ($sourcefs, $targetfs, $snaps) = @_;
+	my ($sourcefs, $targetfs, $snaps, $maxcreation) = @_;
 	foreach my $snap ( sort { $snaps->{'source'}{$b}{'creation'}<=>$snaps->{'source'}{$a}{'creation'} } keys %{ $snaps->{'source'} }) {
+		next if (defined $maxcreation && defined $snaps->{'target'}{$snap} && $snaps->{'target'}{$snap}{'creation'} > $maxcreation);
 		if (defined $snaps->{'target'}{$snap}) {
 			if ($snaps->{'source'}{$snap}{'guid'} == $snaps->{'target'}{$snap}{'guid'}) {
 				return $snap;


pgassmann:
one caveat I found is that there will be an error if there is not yet a newer snapshot with the reduced frequency. it will try to do an incremental from the existing hourly to the daily from yesterday. 

```
INFO: Sending incremental backup@backup4:zfsbulk/backup@autosnap_2026-01-06_12:01:18_hourly ... autosnap_2026-01-06_00:00:02_daily to zfsdata/sync/zfsbulk/backup (~ UNKNOWN):
warning: cannot send 'zfsbulk/backup@autosnap_2026-01-06_00:00:02_daily': not an earlier snapshot from the same fs
```
as this should be a temporary error, I'm ok with that for now.

could certainly be fixed with a nicer warning. what's your opinion on these changes?

pgassmann:
> as this should be a temporary error, I'm ok with that for now.

The added commit fixes the reverse incremental error. it was not just a temporary issue. when the source has stopped receiving updates and the last snapshot was newer than the not filtered ones, e.g. the last daily + some hourly are a week old. it keeps creating errors.

phreaker0:
@pgassmann getsnaps can be quite expensive on hosts with a huge amount of snapshots, it shouldn't be run twice (especiall zfs list) needlessly. It's better to restructure so a single getsnaps call can provide both the filtered and unfiltered list and only calling zfs once so the is no performance hit.

jimsalterjrs:
it's also going to need rebasing, as I'm in the midst of merging a bunch of PRs right now. Sorry 'bout that.

PR #1068, opened 2026-01-17
syncoid: Add --recursive-hold option

brian-maloney:
When running syncoid with `--sendoptions='R'` and `--use-hold`, only the root of the sent hierarchy is held between syncoid runs.  This could result in child datasets being rotated while the root dataset remains.

This PR adds a new --recursive-hold option to syncoid. When specified, it applies the 'zfs hold' and 'zfs release' commands recursively (using the -r flag) on both the source and target datasets.  An alternate option would be to detect the presence of `R` in the `--sendoptions` flag but I think explicit behavior might be better in this case.


phreaker0:
@brian-maloney instead of an extra flag wouldn't it be better to check for the 'R' sendoptions to handle this case automatically? And while at it one can also add a check for conflicts if "--recursive" is given.
What do you think?

jjakob:
This would be very useful. I also think @phreaker0 suggestion about automatically enabling recursive holds is better than an explicit command option, unless there is a good reason that you wouldn't want recursive holds when doing a recursive send.

Issue #1002, opened 2025-05-21
Why `--create-bookmark` implies `--no-sync-snap`?


VasiliPupkin256:
The common backup scenario is:
1) automatically create a new snapshot on the source;
2) incrementally replicate it to the destination, using already existing snapshots and bookmarks;
3) convert the snapshot to the bookmark on the source and leave the snapshot on the destination

This allows to save space on the source, because bookmark doesn't refer any data.

What was the intuition to imply `--no-sync-snap` with `--create-bookmark`? Is it possible to reproduce the incremental bookmark scenario with syncoid?

phreaker0:
- syncoid snapshot is made
- replication is done with syncoid snapshot
- bookmark is created for the syncoid snapshot
- on the next replication the syncoid snapshot is destroyed on source and target
-  created bookmark is useless now

VasiliPupkin256:
Why is it destroyed on target? `--create-bookmark` only converts snapshot to a bookmark and destroys it on the source. Am I missing something?
You can make an incremental backup from the source to the target if there is a bookmark on the source and identical snapshot on the target. This is the point of bookmarks.

phreaker0:
- --create-bookmark will create a bookmark for the last replicated snapshot
- if "--no-sync-snap" is omitted the last replicated snapshot is the one which syncoid creates for this replication
- the old syncoid snapshots are destroyed on each successful replication

VasiliPupkin256:
Yes this is what I want to do. Make syncoid create a new snapshot on the source, replicate it to the target and convert it to the bookmark on the source. Leaving the source without any snapshot and a target with a single last snapshot pointing to the same GUID as the bookmark left on the source. This would allow it to run an incremental backup the next time.

phreaker0:
the syncoid snapshot on the target is also destroyed

VasiliPupkin256:
Why? The manual say:
`--create-bookmark` Creates a zfs bookmark for the newest snapshot on **the source** after replication succeeds

phreaker0:
https://github.com/jimsalterjrs/sanoid/issues/1002#issuecomment-2897091322

VasiliPupkin256:
I've checked the code and it doesn't create a bookmark on the target and doesn't destroy the snapshot on the target. https://github.com/jimsalterjrs/sanoid/blob/6beef5fee67deb2c17f160244953bd5a1983e1ad/syncoid#L831
[...]

Issue #602, opened 2020-12-03
syncoid: Better handle replication when the `--no-rollback` option is specified and bookmarks exist on the source


ein-stein-chen:
The problem:

Infrequently backups with the `--no-rollback` option.  
When the latest snapshot replicated is a daily snapshot it will probably not exist anymore on the `source` when doing the next replication.

So syncoid finds a common snapshot, but can't use it successfully as a base for replication because it is not the latest one on the `target` and it is not allowed to rollback the target.  
(The error is `cannot receive incremental stream: destination … has been modified since most recent snapshot`)

I added the `--create-bookmark` option, so syncoid creates a bookmark for the latest replicated snapshot.  
But syncoid only checks for bookmarks if it can't find any common snapshots between `source` and `target`.

What is possible with the `--no-rollback` option:  
A `send -i` from the bookmark on the source corresponding to the latest snapshot on the target.  
The snapshot to send can be either the latest snapshot on the target (so no intermediate snapshots are transferred)  
or the oldest snapshot which is newer than the bookmark (In this case syncoid could afterward continue with a normal `send -I` from the new snapshot, which would also preserve the intermediate snapshots).

Below is a bash script attached for reproducing the problem.  
It's probably a bit more verbose than necessary.

<details>
  <summary>Click to expand</summary>

```bash
#!/bin/bash

# Cleanup before:
# zfs destroy pool
# rm -r /tmp/zfstest

if [ -d "/tmp/zfstest" ]; then
  echo "Error! /tmp/zfstest directory already exists!"
  exit 1
fi

mkdir /tmp/zfstest

# Latest master version independent from any installed version
curl -s https://raw.githubusercontent.com/jimsalterjrs/sanoid/master/syncoid -o /tmp/zfstest/syncoid
chmod 755 /tmp/zfstest/syncoid

truncate -s 512M /tmp/zfstest/pool
zpool create pool /tmp/zfstest/pool

zfs create pool/a
zfs snapshot pool/a@s0
zfs list -t filesystem,snapshot,bookmark
echo
echo

/tmp/zfstest/syncoid --no-privilege-elevation --no-sync-snap --no-rollback --create-bookmark       pool/a       pool/b
echo
zfs list -t filesystem,snapshot,bookmark
echo
echo

echo "Test 1" > /pool/a/file1
zfs snapshot pool/a@s1
zfs list -t filesystem,snapshot,bookmark
echo
echo

[...]

PR #1074, opened 2026-02-17
Fix bookmark last snapshot sync early exit

### LINKED ISSUE #797: Syncoid: bookmarks not created when latest snapshot is used
Scenario: source gets regularly updated and each update is marked with a new snapshot. The previous snapshot on the source is destroyed afterwards.

syncoid creates the first bookmark on the source, but no more bookmarks afterwards. As a result zfs keeps sending increasingly more data each run as the target gets rolled-back to the target snapshot corresponding to the source bookmark which gets progressively older. IMO it should use a bookmark corresponding to the latest synchronized snapshot instead.

I tracked  the problem to this check
https://github.com/jimsalterjrs/sanoid/blob/ef601eb96b016786af63ea5d7e30c8ad1b5de689/syncoid#L769-L772

which returns from  the syncdataset method before the code creating the bookmarks https://github.com/jimsalterjrs/sanoid/blob/ef601eb96b016786af63ea5d7e30c8ad1b5de689/syncoid#L814 gets called.

ifazk:
Fixes #797

### DIFF
--- a/syncoid
+++ b/syncoid
@@ -726,11 +726,8 @@ sub syncdataset {
 
 			# do a normal replication if bookmarks aren't used or if previous
 			# bookmark replication was only done to the next oldest snapshot
-			if (!$bookmark || $nextsnapshot) {
-				if ($matchingsnap eq $newsyncsnap) {
-					# edge case: bookmark replication used the latest snapshot
-					return 0;
-				}
+			# edge case: skip repilcation if bookmark replication used the latest snapshot
+			if ((!$bookmark || $nextsnapshot) && !($matchingsnap eq $newsyncsnap)) {
 
 				($exit, $stdout) = syncincremental($sourcehost, $sourcefs, $targethost, $targetfs, $matchingsnap, $newsyncsnap, defined($args{'no-stream'}));
 


jimsalterjrs:
Thank you for the patch, ifazk! I think there's still an issue, though; when I tested it on a toy pool, it refused to sync:

````
root@elden:/# zfs list -rt snap source/ds
NAME          USED  AVAIL  REFER  MOUNTPOINT
source/ds@0     0B      -    57K  -
root@elden:/# syncoid-dev --no-sync-snap --create-bookmark source/ds target/ds
WARNING: CRITICAL: no snapshots exist on source source/ds, and you asked for --no-sync-snap.
````

When not using --no-sync-snap, replication succeeds, but the --create-bookmark option appears to be quietly ignored (graned, the help section does explicitly warn you that it only works **with** no-sync-snap):

````
root@elden:/# syncoid-dev --create-bookmark source/ds target/ds
INFO: Sending oldest full snapshot source/ds@syncoid_elden_2026-02-17:11:44:02-GMT-05:00 to new target filesystem target/ds (~ 95 KB):
 127KiB 0:00:00 [18.0MiB/s] [==============================] 134%            
root@elden:/# zfs list -rt all source/ds
NAME                                                    USED  AVAIL  REFER  MOUNTPOINT
source/ds                                                57K   832M    57K  /source/ds
source/ds@0                                               0B      -    57K  -
source/ds@syncoid_elden_2026-02-17:11:44:02-GMT-05:00     0B      -    57K  -
````

**Comment by ifazk** [comment]
That is an unrelated bug that triggers when you name your snapshot `0`... There's a
```perl
if ($newsyncsnap eq 0) {
...
}
```

jimsalterjrs:
> That is an unrelated bug that triggers when you name your snapshot `0`... There's a
> 
> ```perl
> if ($newsyncsnap eq 0) {
> ...
> }
> ```

well shit. Maybe I'll look at that one once I've tested your patch again with an @1 available!

jimsalterjrs:
Thanks again @ifazk , I verified your patch works properly when the snapshot in question is not named `@0`.

````
root@elden:/# zfs snapshot source/ds@1

root@elden:/# syncoid --no-sync-snap --create-bookmark source/ds target/ds
INFO: Sending oldest full snapshot source/ds@1 to new target filesystem target/ds (~ 95 KB):
 127KiB 0:00:00 [9.81MiB/s] [===========================================] 134%            

root@elden:/# cp /usr/local/bin/syncoid-dev /source/ds/
root@elden:/# zfs snapshot source/ds@2

root@elden:/# syncoid --no-sync-snap --create-bookmark source/ds target/ds
INFO: Sending incremental source/ds@1 ... 2 to target/ds (~ 90 KB):
91.0KiB 0:00:00 [3.79MiB/s] [==========================================>] 100%            

root@elden:/# zfs list -rt all source/ds target/ds
NAME          USED  AVAIL  REFER  MOUNTPOINT
source/ds     104K   832M    91K  /source/ds
source/ds@1    13K      -    57K  -
source/ds@2     0B      -    91K  -
source/ds#1      -      -    57K  -
source/ds#2      -      -    91K  -
target/ds     104K   832M    91K  /target/ds
target/ds@1    13K      -    57K  -
target/ds@2     0B      -    91K  -
````

As long as you've been digging around in the code... any idea why --create-bookmark must be used *without* a sync snapshot? 

It seems like using that feature *with* sync snaps--and automatically deleting the source snap once the bookmark has been created and replication has been verified successful--would be an extremely popular way to use this feature.

**Comment by ifazk** [comment]
From what I can tell, that limitation is there so that the following line does not fail. It's used in case there's a naming conflict for naming bookmarks.
```perl
my $guid = $snaps{'source'}{$newsyncsnap}{'guid'};
```
We don't update the `$snaps{'source'}` map after we create a sync snap.

There was also some discussion in https://github.com/jimsalterjrs/sanoid/issues/1002, but I didn't understand it.

The limitation has been there since the beginning of the bookmarks creation feature https://github.com/jimsalterjrs/sanoid/pull/342. We should probably remove the limitation, and just update the `$snaps{'source'}` map.

phreaker0:
@ifazk @jimsalterjrs the snapshot created by syncoid is destroyed on the source and on the target on the next successful replication run, rendering a bookmark created for the syncoid snapshot useless. This was the initial reason for not creating one with "--create-bookmark".

ifazk:
Ah, I get it now. The expectation is that that nothing deletes the sync-snap other than syncoid in the next sync, so create-bookmark is useless when used with sync-snaps because the sync-snap should exist.

From  @jimsalterjrs comment it sounds like it would be popular to create a bookmark on the source for the sync-snap and delete the sync-snap from the source after a successful sync, and on the next sync use the bookmark instead.

For this we also need bookmark syncing to be a bit more first-class than the fallback mechanism that we have now, right now we rollback to the current matching snapshot even if a later bookmark exists, as experienced in https://github.com/jimsalterjrs/sanoid/issues/602.

cc @phreaker0 @jimsalterjrs

jimsalterjrs:
Yeah. It's safe to destroy the bookmarked snapshot on the source immediately after replication completes successfully; we do not *need* to wait for a second replication before doing so.

In fact I think it might make sense for the *default* behavior to be creation of sync snap and bookmark, replication to target, then verify presence of snapshot on target (confirming replication completed successfully), then immediately destroy the snapshot on the source.

This would most closely resemble carefree operation of a tool like rsync for our least clued-in users, which has always been a primary project goal. (Bookmarks didn't EXIST when I first created the project, or I'd have done it that way from the start!)

PR #1090, opened 2026-04-24
[Feature] Prometheus support

deajan:
This PR adds optional Prometheus metrics generation to both sanoid and syncoid, allowing to enhance sanoid/syncoid monitoring.

I had "a no additional perl dependencies developpment" in mind while coding this, this adds only standard lib imports, so existing setups won't be impacted by this change.
Prometheus metrics are collected on success and errors, hence multiple calls to `prom_finalize` so we always get an operational result.

In order to avoid major code rewrites in syncoid, I made a wrapper around `syncdataset()` so changes are kept as minimal as possible.

Metrics include operation status, duration, number of snapshots created/pruned/replicated, thoughput and bytes sent.

This has been running on a couple of my servers for awhile.
I'm open to any suggestions needed for this PR to land in master, so I don't have to maintain a fork.


deajan:
Also, I was not sure whether modifying the README.md file to add docs and grafana dashboard example images (the screenshots come from my setup) is a good idea.

Please let me know if you want to me to remove / rework the readme file additions.

deajan:
I know the project is mature and as such developpment is slow, but I'm pretty sure that's a really helpful feature.
Anything I can produce / test / help with to get a review ?

Issue #1102, opened 2026-06-17
Use of uninitialized value $output_t in pattern match (m//) at /usr/sbin/syncoid line 1450


jjakob:
When the ssh command to run zfs on the remote machine is blocked (the ssh fails), syncoid prints uninitialized value warnings, as well as a critical error about no matching snapshots.
It would be better if it said that the ssh command failed, so that the administrator knew what the problem is.

```
Use of uninitialized value $output_t in pattern match (m//) at /usr/sbin/syncoid line 1450.
Use of uninitialized value $output_p in pattern match (m//) at /usr/sbin/syncoid line 1463.
CRITICAL ERROR:
        Target <redacted> exists but has no snapshots matching with <redacted>!
        Replication to target would require destroying existing
        target. Cowardly refusing to destroy your existing target.
```

https://github.com/jimsalterjrs/sanoid/blob/d39b51a013081e86bb26a6859866c34d98c39ea3/syncoid#L1445

PR #1105, opened 2026-07-22
syncoid: avoid uninitialized value warnings when zfs feature check produces no output

SAY-5:
When the ssh command that probes remote `zfs get` capabilities fails, the pipe returns no output and `$output_t` / `$output_p` stay undefined, so the following pattern matches warn with "Use of uninitialized value in pattern match" (reported in #1102). Defaulting the reads to an empty string keeps the feature detection working (an empty result means the option/property is treated as unsupported) and stops the warnings.


SAY-5:
Ping on this two line one. It silences the uninitialized value warnings syncoid prints when the zfs feature check returns nothing.

